PR #8, opened 2019-09-13
Enhanced printing of components and forecasts.

daniel-thom:


src/utils/print.jl
```diff
@@ -0,0 +1,161 @@
...
+
+function Base.show(io::IO, ::MIME"text/plain", forecasts::Forecasts)
+    println(io, "Forecasts")
+    println(io, "=========")
+    println(io, "Resolution: $(forecasts.resolution)")
+    println(io, "Horizon: $(forecasts.horizon)")
+    println(io, "Interval: $(forecasts.interval)\n")
+    println(io, "---------------------------------")
+    initial_times, dfs = create_forecasts_df(forecasts)
+
```
jd-lara: We should add here the total number of initial_times.

jjstickel:
Reviewing this is probably beyond me at the moment. Will this supersede my PR in PowerSystems.jl? 

I do have one high-level comment. It looks like `Base.summary()` is intended to have two methods, one for IO and another that returns a string (when no IO argument is provided). It looks like you are not providing methods that return the string. (In my PR, I missed providing methods with the IO arument.)

See the output of `?Base.summary`. Quick testing with an array gives
```
julia> summary(stdout, y)
100-element Array{Float64,1}
julia> summary(y)
"100-element Array{Float64,1}"
```
So it looks like the convention is: `summary(obj)` returns a short string and `summary(IO, obj)` is the same as `print(summary(obj))`.

daniel-thom:
Good catch.  I'll fix it.  This does supersede your PR.  We've changed the world on you.  It will hopefully make your life in WaterSystems much easier.

PR #17, opened 2019-09-23
Bump 1.2.3

jd-lara:


CHANGELOG.md
```diff
@@ -1 +1,6 @@
 # InfrastructureSystems
+
+### 0.1.4
```
daniel-thom: What is 0.1.4?

jd-lara: Just a place holder for the next version. I have seen this in other packages. Can eliminate it if you think is best.

daniel-thom: I claim no authority on the subject, but it seems weird to me.

PR #20, opened 2019-10-15
Store forecasts within components and time series data on disk.

daniel-thom:
More changes are still forthcoming, but now is a good time to start reviewing the interfaces.

@jd-lara One thing to focus on is the way the user should iterate over a component's forecasts.  They are stored in a Dict where the key is a Tuple(forecast_type, initial_time, label).  I provided functions to get any of the possibilities, including get_forecast_keys. But what will a user's iteration loop look like?

@kdheepak You should probably pretend this is a new set of code and ignore the diff.


src/component.jl
```diff
@@ -0,0 +1,185 @@
...
+
+function get_forecast_labels(
+                             ::Type{T},
+                             component::InfrastructureSystemsType,
+                             initial_time::Dates.DateTime,
+                            ) where T <: Forecast
+    return get_forecast_labels(forecast_external_to_internal(T), _get_forecast_container(component), initial_time)
+end
+
+function get_forecast_value(
```
jd-lara: It will be better if we return 

`get_timeseries(forecast)[index]*get_field(component, label)`

src/component.jl
```diff
@@ -0,0 +1,185 @@
...
+end
+
+function get_time_series_uuids(component::InfrastructureSystemsType)
+    container = _get_forecast_container(component)
+
+    return [(get_time_series_uuid(container.data[key]), key.label)
+             for key in get_forecast_keys(component)]
+end
+
+function iterate_forecasts(component::InfrastructureSystemsType)
```
jd-lara: We might want something like this that iterates over the forecasts with the same initial time, not the forecasts inside of a component. Unless is used internally for something else, I don't think we would need this feature.

src/components.jl
```diff
@@ -302,3 +314,49 @@ function get_num_components(components::Components)
...
+    initial_times = Set{Dates.DateTime}()
+    for component in iterate_components_with_forecasts(components)
+        get_forecast_initial_times!(initial_times, component)
+    end
+
+    return sort!(Vector{Dates.DateTime}(collect(initial_times)))
+end
+
+function get_forecasts_initial_time(components::Components)
+    initial_times = get_forecast_initial_times(components)
```
jd-lara: it's better to throw an error here or an exception.

src/components.jl
```diff
@@ -302,3 +314,49 @@ function get_num_components(components::Components)
...
+    return isempty(initial_times) ? nothing : initial_times[1]
+end
+
+function get_forecasts_last_initial_time(components::Components)
```
jd-lara: same here

src/forecasts.jl
```diff
@@ -1,443 +1,144 @@
...
+"""
+Abstract type for forecasts supplied to users. They are not stored in a system. Instead,
+they are generated on demand for the user.
+Users can create them. The system will convert them to a subtype of ForecastInternal for 
+storage.
+Time series data is stored as a field, so reads will always be from memory.
+"""
+abstract type Forecast <: Any end
 
+get_horizon(forecast::Forecast) = length(get_data(forecast))
```
jd-lara: if get `get_data` returns the whole time_series then the results won't be consistent with the internal horizon value?

src/supplemental_accessors.jl
```diff
@@ -1,16 +1,19 @@
-"""Get Forecast component name."""
-get_forecast_component_name(forecast::Forecast) = get_name(get_component(forecast))
+"""
+    get_forecast_value(val::Deterministic, ix)
 
-"""gets the value of a Deterministic forecast at a given index or DateTime timestamp"""
+Return the value of a Deterministic forecast at a given index or DateTime timestamp.
+"""
```
jd-lara: I think this is no longer needed given the implementation of get_forecast_value in line 130 of Forecasts.jl

src/supplemental_constructors.jl
```diff
@@ -1,91 +1,123 @@
...
-    start_index = 1
-    horizon = length(data)
-    return Deterministic(component, label, resolution, initial_time, data, start_index,
-                         horizon, InfrastructureSystemsInternal())
+function make_public_forecast(forecast::DeterministicInternal, data::TimeSeries.TimeArray)
+    return Deterministic(get_label(forecast), data)
 end
 
-"""Constructs Probabilistic after constructing a TimeArray from initial_time and time_steps.
+function make_internal_forecast(forecast::Deterministic, ts_data::TimeSeriesData)
```
jd-lara: if internal forecasts are not meant to be manipulated by users, shouldn't this be internal.

src/supplemental_constructors.jl
```diff
@@ -1,91 +1,123 @@
...
                        data::TimeSeries.TimeArray)
-    start_index = 1
-    horizon = length(data)
-    return Probabilistic(component, label, resolution, initial_time, percentiles, data,
-                         start_index, horizon, InfrastructureSystemsInternal())
+    return Probabilistic(label, percentiles, data)
+end
+
+
+function make_public_forecast(forecast::ProbabilisticInternal, data::TimeSeries.TimeArray)
```
jd-lara: same here

src/system_data.jl
```diff
@@ -377,31 +423,154 @@ end
...
+                                    component::InfrastructureSystemsType
+                                   ) where T <: Forecast
+    forecast_type = forecast_external_to_internal(T)
+    return get_forecast_initial_times(forecast_type, component)
+end
+
+function get_forecasts_interval(data::SystemData)
+    initial_times = get_forecast_initial_times(data)
+    if length(initial_times) <= 1
+        return UNINITIALIZED_PERIOD
```
jd-lara: Why return UNINITIALIZED_PERIOD?

src/component.jl
daniel-thom: For debug purposes I commonly do something like this:
```
forecasts = collect(iterate_forecasts(component))
```

src/forecasts.jl
daniel-thom: I think this was resolved in our conversation.

src/supplemental_accessors.jl
daniel-thom: You can call this function if you have an instance of Forecast.  The function on line 130 of component.jl allows you to call it without an instance of a forecast.  Both seem useful.  Are you sure you want to remove this one?

```
130 function get_forecast_value(
131                             ::Type{T},
132                             component::InfrastructureSystemsType,
133                             initial_time::Dates.DateTime,
134                             label::AbstractString,
135                             index::Union{Int, Dates.DateTime},
136                            ) where T <: ForecastInternal
```

src/supplemental_constructors.jl
daniel-thom: Not sure what you mean.  The function is not exported.  If you are referring to lack of a leading underscore, my personal policy is to lead with an underscore if a function is only used within a file (also a Python convention).  I may not have always strictly followed that, though.

src/system_data.jl
daniel-thom: UNINITIALIZED_PERIOD = 0 seconds.  The interval is defined as the difference between initial times.  In order to have an interval greater than 0 you would have to have at least two initial times.  I suppose it would be better to use 0 explicitly here.

src/component.jl
daniel-thom: @claytonpbarrows @jd-lara 
The process that consumes the user descriptor file that maps user CSV column names to PowerSystem names is implemented in the PowerSystemTableData struct.  It is used during parsing but not normal operation.

Options:

1. Convert label during parsing only.  This means that any code that creates forecasts or calls add_forecast!(sys) would have to change the label being passed.
2. Restructure the code such that the System consumes the descriptors and performs the conversion.

I prefer #1 because it makes better sense to me and is much simpler.

jd-lara: Also prefer 1.

src/supplemental_accessors.jl
jd-lara: ok

src/supplemental_constructors.jl
jd-lara: I have tried to implement that all non-exported functions carry the _ in the beginning for clarity.

src/system_data.jl
jd-lara: ok

daniel-thom: Another important note:  we've never had code that verifies intervals are consistent. I'm wondering if we should remove the concept from the System.  It could only exist when you're generating 'split' forecasts on the fly.

jd-lara: I prefer to validate it when creating the forecast object so we support using forecasts outside of the system.

src/supplemental_constructors.jl
daniel-thom: We should have a rule for team consistency. A leading underscore is common for many people if the function is in the same file.  I find it a bit distracting if it's jumping across files.  Judging whether it's public requires an extra search, but that's ok with me.  That is very similar to other languages.  Python is the only language I know of where the underscore is the telling factor.

src/component.jl
claytonpbarrows: 1 sounds good

daniel-thom:
This actually points to a bigger problem with attaching forecasts to subtypes of TechnicalParams.  They aren't subtypes of PSY.Component.  You will not get as far as this function.  The function the user calls is PSY.add_forecast!().  That function signature requires a subtype of Component.  The current code in master requires that a Forecast have a Component.

It shouldn't be a big change to make this more broad.  But I think there are plenty of functions in parsing land that rely on the rule that a Component implements get_name().

jd-lara:
We can have generic get_names for components that return their names so we don't name every component.

daniel-thom:
I don't think we should call get_name() on any object that doesn't have a name.

jd-lara:
Then we shouldn't rely on `get_name` for the forecast to be attached.

daniel-thom:
That's a non-trivial change that we just identified.  I suggest that we get through the existing work and then think about what to do.  All of the parsing code assumes that the time series column name is the component name.

jd-lara:
We can't generate tests systems without this change because we can't attach data to the devices

daniel-thom:
The previous design didn't allow for it, either.  I'm just trying to avoid feature creep to get this work done.

src/component.jl
daniel-thom: @jd-lara Would it be better to do this multiplication on the TimeArray as soon as we add the forecast to the component?  Doing what you suggested could mean that the user gets different values depending on what function they call.

jd-lara: I don't think that doing the multiplication is the best option since the idea is that the user can manipulate the value that the scaling factor is being multiplied by. As for the consistency, we can change the naming to `get_scalingfactors` and return the time series and `get_forecast_value` that returns the value at the time for the label. This will avoid confusion.

daniel-thom: There is some nuance to this.  If the user calls get_forecast() then they will receive a Forecast object that has scaling factors (because we store scaling factors).  There is a get_forecast_value method on that object which just returns a value at an index - so that would return a scaling factor.  (But we could delete that method.)

If the user calls the get_forecast_value method that takes a Component then would could perform the multiplication.

Another option is to add an optional parameter to get_forecast to allow specification of scaling_factor or not.  Then the Forecast object they get back would have exactly what they want.

I think I like that best.  Here is what the functions would look like.  Let me know what you think.
```
get_forecast(T, sys, component, initial_time, label; scaling_factor=true)
get_forecast_value(T, sys, component, initial_time, label, index; scaling_factor=true)
get_forecast_value(forecast::Forecast, index)
get_data(forecast::Forecast) = forecast.data
get_time_series(forecast::Forecast) = get_data(forecast)
```

src/component.jl
```diff
@@ -0,0 +1,385 @@
...
+    throw(ArgumentError("did not find a forecast matching the requested parameters"))
+end
+
+"""
+    get_forecast_values(component::InfrastructureSystemsType, forecast::Forecast)
+
+Return a TimeSeries.TimeArray where the forecast data has been multiplied by the forecasted
+component field.
+"""
+function get_forecast_values(component::InfrastructureSystemsType, forecast::Forecast)
```
daniel-thom: @jd-lara Can you confirm that this is what you want?

jd-lara: No, this is what I am looking for 

```Julia
function get_forecast_values(component::InfrastructureSystemsType, label)
```

daniel-thom: If you don't want to pass the forecast object then it would have to be
```
function get_forecast_values(::Type{T<:Forecast}, component::InfrastructureSystemsType, initial_time, label)
```

PR #22, opened 2019-10-21
Auto-delete temporary HDF5 files.

daniel-thom:


src/hdf5_time_series_storage.jl
```diff
@@ -5,22 +5,35 @@ const HDF5_TS_ROOT_PATH = "time_series"
...
 """
 Constructs Hdf5TimeSeriesStorage by creating a new file.
 """
-function Hdf5TimeSeriesStorage(; filename=nothing)
-    file_path = isnothing(filename) ? tempname() * ".h5" : filename
-    storage = Hdf5TimeSeriesStorage(file_path)
+function Hdf5TimeSeriesStorage()
+    storage = Hdf5TimeSeriesStorage(tempname() * ".h5")
     _make_file(storage)
-    @debug "Created time series storage file." storage.file_path
```
jd-lara: This is a useful message to keep.

daniel-thom: Not anymore. It will get deleted after the system gets garbage collected.

PR #27, opened 2019-11-05
Change forecast labels to component accessor functions

daniel-thom:


daniel-thom:
@jd-lara This should now be final.

PR #32, opened 2019-11-15
Fix get_forecast for contiguous time series arrays

daniel-thom:


daniel-thom:
@Lilyhanig I added code and tests to cover the missing case.  Can you test this code with PowerSimulations?

daniel-thom:
Lily says that PowerSimulations is working with this patch.

PR #35, opened 2019-11-20
Throw error on add_forecast for unsupported component

daniel-thom:


src/component.jl
```diff
@@ -2,7 +2,12 @@ function add_forecast!(
                        component::T,
                        forecast::ForecastInternal,
                       ) where T <: InfrastructureSystemsType
-    add_forecast!(_get_forecast_container(component), forecast)
+    container = _get_forecast_container(component)
+    if isnothing(container)
+        throw(ArgumentError("type $T does not contain forecasts"))
```
jd-lara: The message isn't clear if the cause of the error is the type $T not having the forecast field or the data not being available. 

I would suggest 
```Julia
throw(ArgumentError("type $T can not store forecasts"))
```

daniel-thom: Fixed

src/component.jl
```diff
@@ -4,7 +4,7 @@ function add_forecast!(
                       ) where T <: InfrastructureSystemsType
     container = _get_forecast_container(component)
     if isnothing(container)
-        throw(ArgumentError("type $T does not contain forecasts"))
+        throw(ArgumentError("type $T cannot store forecasts"))
```
jd-lara: ```throw(ArgumentError("type $T does not support storing forecasts"))``` ?

daniel-thom: That's better.  Fixed.

PR #36, opened 2019-11-21
Jd/add ext

jd-lara:
This PR adds the additional constructor when ext is included in the structs definition. I added to the PR a change in the way internals is printed to enable dynamics.

### DIFF
--- a/src/generated/Deterministic.jl
+++ b/src/generated/Deterministic.jl
@@ -20,9 +20,12 @@ mutable struct Deterministic <: Forecast
     data::TimeSeries.TimeArray
 end
 
+
 function Deterministic(; label, data, )
     Deterministic(label, data, )
 end
+
+
 
 """Get Deterministic label."""
 get_label(value::Deterministic) = value.label
--- a/src/generated/Deterministic.jl
+++ b/src/generated/Deterministic.jl
@@ -20,14 +20,9 @@ mutable struct Deterministic <: Forecast
     data::TimeSeries.TimeArray
 end
 
-
-
 function Deterministic(; label, data, )
     Deterministic(label, data, )
 end
-
-
-
 
 """Get Deterministic label."""
 get_label(value::Deterministic) = value.label
--- a/src/generated/Deterministic.jl
+++ b/src/generated/Deterministic.jl
@@ -5,6 +5,7 @@ This file is auto-generated. Do not edit.
     mutable struct Deterministic <: Forecast
         label::String
         data::TimeSeries.TimeArray
+        ext::Union{Nothing, Dict{String, Any}}
     end
 
 A deterministic forecast for a particular data field in a Component.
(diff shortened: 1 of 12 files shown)


src/utils/generate_structs.jl
```diff
@@ -104,6 +114,16 @@ function generate_structs(directory, data::Vector; print_results=true)
                 continue
             end
 
+            if param["name"] == "ext"
```
daniel-thom: You could make this an if/elseif/elseif block.

src/utils/generate_structs.jl
```diff
@@ -53,6 +53,16 @@ function {{struct_name}}{T}({{#parameters}}{{^internal}}{{name}}{{#default}}={{d
     {{struct_name}}({{#parameters}}{{^internal}}{{name}}, {{/internal}}{{/parameters}}InfrastructureSystemsInternal())
 end
 {{/parametric}}
+
+{{#defines_ext}}
+function {{struct_name}}({{#parameters}}{{^internal}}{{^ext}}{{^_forecasts}}{{name}}, {{/_forecasts}}{{/ext}}{{/internal}}{{/parameters}}; ext={{#parameters}}{{#ext}}{{default}}{{/ext}}{{/parameters}})
+    {{#parameters}}
+    {{/parameters}}
```
daniel-thom: Isn't this unnecessary?

src/utils/generate_structs.jl
```diff
@@ -53,6 +53,16 @@ function {{struct_name}}{T}({{#parameters}}{{^internal}}{{name}}{{#default}}={{d
     {{struct_name}}({{#parameters}}{{^internal}}{{name}}, {{/internal}}{{/parameters}}InfrastructureSystemsInternal())
 end
 {{/parametric}}
+
+{{#defines_ext}}
+function {{struct_name}}({{#parameters}}{{^internal}}{{^ext}}{{^_forecasts}}{{name}}, {{/_forecasts}}{{/ext}}{{/internal}}{{/parameters}}; ext={{#parameters}}{{#ext}}{{default}}{{/ext}}{{/parameters}})
```
daniel-thom: Why did you have to do something with forecasts?  Seems like the existing logic with 'default' should have taken care of it.

src/utils/generate_structs.jl
jd-lara: I will test it, I based it from the code in line 51.

src/utils/generate_structs.jl
daniel-thom: You could remove the `defines_ext` block if you define ext before _forecast in the descriptors.

jd-lara: I added this to avoid printing the _forecast argument.

src/generated/Deterministic.jl
```diff
@@ -27,6 +27,8 @@ function Deterministic(; label, data, )
 end
 
 
+
```
daniel-thom: There are a bunch of extra blank lines from something in the Mustache template.

src/utils/generate_structs.jl
```diff
@@ -104,6 +96,15 @@ function generate_structs(directory, data::Vector; print_results=true)
...
 
+            if param["name"] == "ext"
+                param["ext"] = true
+                continue
+            end
+
+            if param["name"] == "_forecasts"
+                param["_forecasts"] = true
+            end
+
```
daniel-thom: This is no longer needed, right?

src/utils/generate_structs.jl
```diff
@@ -35,21 +35,14 @@ mutable struct {{struct_name}}{{#parametric}}{T <: {{parametric}}}{{/parametric}
     end
     {{/inner_constructor_check}}
 end
-
```
daniel-thom: Did you mean to remove these blank lines?  Now the functions in the generated code have no whitespace between them.

PR #42, opened 2019-12-11
adding exceptions and overloading the division method

Lilyhanig:
- adding exceptions for Power Simulations
- overloading base division method to accept different time periods


Lilyhanig:
@daniel-thom can you take a look at this?

src/common.jl
```diff
@@ -11,3 +11,11 @@ end
...
     msg::AbstractString
 end
+
+struct ConflictingInputs <: Exception
+    msg::AbstractString
+end
+
+struct HashMismatch <: Exception
+    msg::AbstractString
+end
```
daniel-thom: There is a longstanding convention that each file should end in a newline.  Git complains if you don't have it.  Looks like you took it away here.  Can you add it back?  If you're using VSCode then you can do this automatically with `files.insertFinalNewline": true` in your config settings.

src/units.jl
```diff
@@ -0,0 +1 @@
+timeperiod_conversion(time_period::Dates.TimePeriod) = convert(Dates.Millisecond, time_period)
```
daniel-thom: Shouldn't have both "time_period" and "timeperiod".  "time_period" is better.

src/utils/print.jl
```diff
@@ -124,6 +124,59 @@ function create_components_df(components::Components)
     return DataFrames.DataFrame(rows)
 end
 
+
+function Base.show(io::IO, ::MIME"text/plain", period::Union{Dates.TimePeriod, Dates.DatePeriod})
+    period = time_period_conversion(period)
+
+    milli_weeks = period - (period % Dates.Millisecond(604800000))
+    weeks = convert(Dates.Week, milli_weeks)
+    period = period - milli_weeks
```
daniel-thom: For all of these it would be a bit cleaner to use `period -=`

src/utils/print.jl
```diff
@@ -124,6 +124,59 @@ function create_components_df(components::Components)
...
+    hours = convert(Dates.Hour, milli_hours)
+    period = period - milli_hours
+
+    milli_minutes = period - (period % Dates.Millisecond(60000))
+    minutes = convert(Dates.Minute, milli_minutes)
+    period = period - milli_minutes
+
+    seconds = period - (period % Dates.Millisecond(1000)) # finding the seconds
+    seconds = convert(Dates.Second, seconds)
+    remainder = period % Dates.Millisecond(1000) #finding the remainding milliseconds
```
daniel-thom: All of this logic is duplicated.  If we really need HTML output then I suggest making a helper function.

src/utils/print.jl
```diff
@@ -124,6 +124,59 @@ function create_components_df(components::Components)
...
+    println(io, "$(weeks + days + hours + minutes + seconds + remainder)")
```
daniel-thom: Many people will be surprised by what this does if they only read it.  I suggest adding a comment saying that it will create a compound object and print that.

src/utils/print.jl
jd-lara: This could also be made into a function?

PR #44, opened 2019-12-13
Add redirection of stdout to log events during external function execution

daniel-thom:
The point of this is to allow logging of solver output in PowerSimulations.  I also added code to report timestamps and process IDs in log files.


daniel-thom:
Here is what the code looks like in PowerSimulations.  I'll make a PR once this gets approved and merged.

https://github.com/daniel-thom/PowerSimulations.jl/commit/08d64b7c7e489fbbc827d24ce28273efd1eec3d2

daniel-thom:
@claytonpbarrows The build is failing because that is running Julia 1.1 and I added code that requires 1.2.  I thought we had upgraded.  Can we require our users to be at 1.2 or greater?  If we do need to work at 1.1 then there are ways I can fix it.

jd-lara:
We don't support 1.1 anymore. We need to update the Travis code in IS.

PR #45, opened 2019-12-20
add conversion for time period dictionaries

Lilyhanig:


daniel-thom:
The changes look good.  Can you add a test?

PR #47, opened 2020-01-14
add valid range to doc string

Lilyhanig:


jd-lara:
@daniel-thom we can merge this as is. It will require a PR in PSY to update all the structs. I can take care of that.

PR #53, opened 2020-02-01
Jd/formatter fix

jd-lara:
There was an issue with the formatter option. This PR fixes it


jd-lara:
closed in favor of #54

PR #56, opened 2020-02-06
Ignore line endings when comparing generated structs

daniel-thom:
This fixes failures on Travis Windows systems.  It also allows other packages to use the same code for validating generated structs.


daniel-thom:
@jd-lara Once this gets merged and tagged, I have PowerSystems changeset ready to go.

PR #57, opened 2020-02-07
Allow auto-generated accessor functions to be defined in another module

daniel-thom:
This will allow us to fix https://github.com/NREL/PowerSystems.jl/issues/422.


jd-lara:
@daniel-thom you can merge as soon as Travis passes

PR #62, opened 2020-02-20
add results abstract type

Lilyhanig:


src/results.jl
```diff
@@ -0,0 +1 @@
+abstract type Results end
```
daniel-thom: I think this is too abstract.  We should define the behavior of this object.  This could be by defining interface functions or fields.

PR #65, opened 2020-02-25
Forecast field change

daniel-thom:


src/InfrastructureSystems.jl
```diff
@@ -17,6 +17,14 @@ abstract type InfrastructureSystemsType end
 
 get_name(value::InfrastructureSystemsType) = value.name
 
+"""
+Return the internal forecast storage container or nothing, if the type doesn't store
+forecasts.
+
+Subtypes need to implement this method if they store forecasts.
+"""
+get_forecasts(value::InfrastructureSystemsType) = nothing
```
daniel-thom: Note that this will require a change from anyone who consumes IS.  I could put a workaround to check for the existence of the field `_forecasts`.

PR #66, opened 2020-03-04
adding data file with DateTime column

claytonpbarrows:
This PR adds a test datafile with a "DateTime" column, but no "Period" column, modifies the detection of the file format, and handles the lack of `file.Period` in the calculation of `step`.


daniel-thom:
The old code (which didn't fully support DateTime since there were no tests) was written assuming that this format would have both DateTime and Period as columns.  Hence it is was called TimeseriesFormatDateTimePeriodAsColumn and was a subtype of TimeseriesFormatPeriodAsColumn.  That is now very misleading.

I think we should change this format type to TimeseriesFormatDateTimeAsColumn and make it a subtype of TimeseriesFileFormat.  That would require another minor change to the function `get_step_time`.  I'll volunteer for the new work.  I think we should do it before merging.

src/timeseries_formats.jl
```diff
@@ -41,6 +42,8 @@ function get_timeseries_format(file::CSV.File)
         end
     elseif has_ymd
         format = TimeseriesFormatYMDPeriodAsHeader
+    elseif has_datetime
+        format = TimeseriesFormatDateTimeAsColumn
```
daniel-thom: The line below this can now be changed to `else`.

PR #71, opened 2020-04-04
PSIMA-65: Event Recorder

daniel-thom:
This PowerSimulations pull request shows how this feature can be used.:  https://github.com/NREL-SIIP/PowerSimulations.jl/pull/406.

Refer to the tests in test_simulation_execute.jl for specific examples of the entire use case.


src/utils/recorder_events.jl
```diff
@@ -0,0 +1,290 @@
...
+
+    return T(vals...)
+end
+
+"""
+Records user-defined events in JSON format.
+"""
+mutable struct Recorder
+    name::Symbol
+    io::IO
```
daniel-thom, line 63: I considered storing a subtype of Logging.AbstractLogger instead of an IO.  I decided against that because we don't need the benefits of the logger here.
- Controlled whether a particular recorder is enabled or not is done with the register_recorder! function.  The "log level" will always be the same.
- The default logger adds things like file and line number to the message automatically.  We would have to write code to prevent that.

It would be easy to switch to a logger if you have arguments for it.

src/utils/recorder_events.jl
```diff
@@ -0,0 +1,290 @@
...
+- `directory = "."`:  Only used when io is nothing.
+"""
+function Recorder(
+    name::Symbol;
+    io::Union{Nothing, IO} = nothing,
+    mode = "w",
+    directory = ".",
+)
+    if isnothing(io)
+        filename = joinpath(directory, string(name) * "_recorder.log")
```
daniel-thom, line 89: The default behavior is store events for each recorder in its own file.  I considered storing all events for all loggers in one file.  My initial thought was that this provides the user better control. Also, it would prevent files with tons of events from crowding out files with few events.  We could add file rotation based on size in the future, if needed.

Feel free to make a different argument.

test/Project.toml
```diff
@@ -14,4 +13,11 @@ UUIDs = "cf7118a7-6976-5b1a-9a39-7adc72f591a4"
 YAML = "ddb6d928-2868-570f-bddf-ab3f9cf99eb6"
 
 [compat]
+CSV = "~0.5"
```
jd-lara, line 16: There might conflict with the main project.toml

PR #83, opened 2020-05-12
Support moving serialization files to new path

daniel-thom:
Fixes issue PowerSystems issue 472.  Requires changes in PSY.

### DIFF
--- a/src/component.jl
+++ b/src/component.jl
@@ -88,7 +88,7 @@ function get_forecast(
         throw(ArgumentError("no forecasts are stored in $component"))
     end
 
-    first_forecast = iterate(iterate_forecasts(component))[1]
+    first_forecast = iterate(iterate_forecasts(ForecastInternal, component))[1]
     resolution = get_resolution(first_forecast)
     sys_horizon = get_horizon(first_forecast)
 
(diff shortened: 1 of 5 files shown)


src/component.jl
```diff
@@ -440,10 +440,30 @@ function prepare_for_removal!(component::InfrastructureSystemsType)
...
 end
 
+"""
+Returns an iterator of Forecast instances attached to the component.
+"""
 function iterate_forecasts(component::InfrastructureSystemsType)
     container = get_forecasts(component)
     forecast_keys = sort!(collect(keys(container.data)), by = x -> x.initial_time)
 
+    Channel() do channel
```
daniel-thom, line 450: This is a secondary issue Sourabh saw today.  This needs to return public forecasts, not internal ones.

src/system_data.jl
```diff
@@ -71,22 +70,36 @@ function SystemData(
...
+    cd(dirname(filename))
+    try
+        return open(filename) do io
+            from_json(io, SystemData)
+        end
+    finally
+        cd(orig_dir)
+    end
+end
+
```
jd-lara, line 102: Are de-serializing from a folder or a JSON file? If from a folder shouldn't we check that the input is a folder?

jd-lara:
I merged this PR and the respective PSY one to be able to do testing with all the code in master. As soon as it passes I can tag.

PR #89, opened 2020-05-15
enable passing abstract type to get component

jd-lara:
This adds a method to get components by name passing abstract types

### DIFF
--- a/src/components.jl
+++ b/src/components.jl
@@ -166,8 +166,8 @@ end
                   name::AbstractString
                  )::Union{T, Nothing} where T <: InfrastructureSystemsType
 
-Get the component of type T with name. Returns nothing if no component matches. If an Abastract
-    Type T is passed the names should be unique accross Abastract Types
+Get the component of type T with name. Returns nothing if no component matches. If T is an abstract
+type then the names of components across all subtypes of T must be unique.
 
 See [`get_components_by_name`](@ref) if the concrete type is unknown.
 
--- a/src/components.jl
+++ b/src/components.jl
@@ -178,7 +178,11 @@ function get_component(
     name::AbstractString,
 )::Union{T, Nothing} where {T <: InfrastructureSystemsType}
     if !isconcretetype(T)
-        throw(ArgumentError("get_component only supports concrete types: $T"))
+        components = get_components(T, components, x -> get_name(x) == name)
+        if length(components) > 1
+            throw(ArgumentError("More than one components of type $T"))
+        end
+        return first(components)
     end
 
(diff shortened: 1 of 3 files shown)


src/components.jl
```diff
@@ -178,10 +178,7 @@ function get_component(
     name::AbstractString,
 )::Union{T, Nothing} where {T <: InfrastructureSystemsType}
     if !isconcretetype(T)
-        components = get_components(T, components, x -> get_name(x) == name)
-        if length(components) > 1
-            throw(ArgumentError("More than one components of type $T"))
-        end
+        components = get_components_by_name(T, components, name)
```
daniel-thom, line 181: The docstring is now incorrect.

More importantly, this could cause very confusing behaviors for our users.  If they call `get_component(Component)` and there are duplicate names across the concrete types, this will return the first one, which will be determined seemingly randomly be the way `get_components` iterates over component types.  Users will not be able to predict what they will get back.

With the current approach, if the user doesn't know the concrete type then they have to call `get_components_by_name` and then check how many components they received.  That might be a bit cumbersome, but it's easy to understand and allows us to convey clear, consistent rules.

I think we should have a group discussion before merging this functionality.

jd-lara: Concrete types don't allow for duplicate names. I understand the concern, but with the current approach, it is way too cumbersome to make models based on names when mixes of types are involved. For instance, setting a reserve value where the contributing devices can have multiple types. 

The objective is to make a simple function that can integrate well with a JuMP macro to define a variable/constraint. Having to do this outside is adding extra overhead for a corner case that would fail anyway if the user has repeated names in the definition of the set.

src/components.jl
```diff
@@ -166,19 +166,25 @@ end
                   name::AbstractString
                  )::Union{T, Nothing} where T <: InfrastructureSystemsType
 
-Get the component of concrete type T with name. Returns nothing if no component matches.
+Get the component of type T with name. Returns nothing if no component matches. If an Abastract
```
daniel-thom, line 169:
```
Get the component of type T with name. Returns nothing if no component matches. If T is an abstract
type then the names of components across all subtypes of T must be unique.
```

src/components.jl
```diff
@@ -166,19 +166,25 @@ end
...
+    Type T is passed the names should be unique accross Abastract Types
 
 See [`get_components_by_name`](@ref) if the concrete type is unknown.
```
daniel-thom, line 172: This should be re-worded.  Maybe
```
See [`get_components_by_name`](@ref) for abstract types with non-unique names across subtypes.
```

PR #90, opened 2020-05-20
PSIMA142: Adding Forecast for ORDC

sourabhdalvi:
- Adding CostCoefficient forecast that stores the breakpoints for piecewise linear cost function of ReserveDemandCurve


src/descriptors/structs.json
```diff
@@ -177,5 +177,68 @@
       }
     ],
     "supertype": "Forecast"
+  },
+  {
+    "struct_name": "CostCoefficient",
```
jd-lara, line 182: We should make this more generic. This should be something like a PWL function forecast.

src/generated/CostCoefficient.jl
```diff
@@ -0,0 +1,44 @@
+#=
+This file is auto-generated. Do not edit.
+=#
+"""
+    mutable struct CostCoefficient <: Forecast
```
jd-lara, line 5: Same Here.

test/test_forecasts.jl
```diff
@@ -769,6 +769,36 @@ end
...
+
+    IS.add_forecast!(sys, component, forecast)
+    forecast2 = IS.get_forecast(IS.CostCoefficient, component, dates[1], label)
+    @test forecast2 isa IS.CostCoefficient
+    fdata2 = IS.get_data(forecast2)
+    @test length(TimeSeries.colnames(fdata2)) == 4
+    @test TimeSeries.timestamp(ta) == TimeSeries.timestamp(fdata2)
+    @test TimeSeries.values(ta) == TimeSeries.values(fdata2)
+end
+
```
daniel-thom, line 801: Can you add tests for `get_forecast` with horizon so that the concatenation code is tested for both CostCoefficient and ScenarioBased?

src/descriptors/structs.json
```diff
@@ -177,5 +177,68 @@
...
+    "docstring": "A forecast for cost function data field in a Component.",
+    "fields": [
+      {
+        "name": "label",
+        "data_type": "String",
+        "comment": "label of component parameter forecasted"
+      },
+      {
+        "name": "break_points",
+        "data_type": "Int64",
```
daniel-thom, line 192: There is inconsistency in our code base on Int and Int64.  If you don't actually care about the number of bits (in this case you almost certainly don't), just use Int and let Julia use the native type for the system.

src/descriptors/structs.json
```diff
@@ -177,5 +177,68 @@
...
+        "name": "data",
+        "data_type": "TimeSeries.TimeArray",
+        "comment": "timestamp - scalingfactor"
+      }
+    ],
+    "supertype": "Forecast"
+  },
+  {
+    "struct_name": "CostCoefficientInternal",
+    "docstring": "A Discrete Scenario Based forecast for a particular data field in a Component.",
```
daniel-thom, line 205: stale copy/paste

src/supplemental_constructors.jl
```diff
@@ -160,6 +160,23 @@ function ScenarioBased(
     return ScenarioBased(label, data)
 end
 
+function ScenarioBased(forecasts::Vector{ScenarioBased})
```
daniel-thom, line 163: Can you modify the comment on line 43?

src/supplemental_constructors.jl
```diff
@@ -160,6 +160,23 @@ function ScenarioBased(
...
+    @assert !isempty(forecasts)
+    scenario_count = get_scenario_count(forecasts[1])
+    colnames = TimeSeries.colnames(get_data(forecasts[1]))
+    data = Array{Any}(undef, 0, scenario_count)
+    timestamps =
+        collect(Iterators.flatten((TimeSeries.timestamp(get_data(x)) for x in forecasts)))
+    for x in forecasts
+        data = vcat(data, TimeSeries.values(get_data(x)))
+    end
```
daniel-thom, line 172:
```suggestion
    data = vcat((TimeSeries.values(get_data(x)) for x in forcasts))
```

src/supplemental_constructors.jl
```diff
@@ -175,13 +192,77 @@ function make_internal_forecast(forecast::ScenarioBased, ts_data::TimeSeriesData
...
+end
+
+function CostCoefficient(forecasts::Vector{CostCoefficient})
+    @assert !isempty(forecasts)
+    break_points = get_break_points(forecasts[1])
+    colnames = TimeSeries.colnames(get_data(forecasts[1]))
+    data = Array{Any}(undef, 0, break_points)
+    timestamps =
+        collect(Iterators.flatten((TimeSeries.timestamp(get_data(x)) for x in forecasts)))
+    for x in forecasts
```
daniel-thom, line 232: same simplification as above

src/descriptors/structs.json
sourabhdalvi: `PiecewiseCost` ?

src/descriptors/structs.json
```diff
@@ -179,8 +179,8 @@
     "supertype": "Forecast"
   },
   {
-    "struct_name": "CostCoefficient",
-    "docstring": "A forecast for cost function data field in a Component.",
+    "struct_name": "PiecewiseCost",
```
jd-lara, line 182: This works for any `PieceWiseFunction` not just cost.

sourabhdalvi: will do

PR #91, opened 2020-05-21
Support storing column names in HDF

daniel-thom:


src/hdf5_time_series_storage.jl
```diff
@@ -164,8 +177,14 @@ function get_time_series(
             timestamps = path["timestamps"][index:end_index]
```
sourabhdalvi, line 177: But you would still get a error here `data = path["data"][index:end_index]` as your passing wrong number of indices to the HDF5 Dataset

daniel-thom: In general that would probably be a good idea.  In this case the forecast controls its own length, so it will never ask for something bigger than what exists.

PR #96, opened 2020-05-30
Jd/add forward macro

jd-lara:
This PR adds the @forward macro capability to IS. The code can't be tested here because IS won't export any method so the tests for this will be in PSY.

### DIFF
--- a/src/component.jl
+++ b/src/component.jl
@@ -393,16 +393,41 @@ end
 """
 Efficiently add all forecasts in one component to another by copying the underlying
 references.
+
+# Arguments
+- `src::InfrastructureSystemsType`: Source component
+- `dst::InfrastructureSystemsType`: Destination component
+- `label_mapping::Dict = nothing`: Optionally map src labels to different dst labels.
+  If provided and src has a forecast with a label not present in label_mapping, that
+  forecast will not copied. If label_mapping is nothing then all forecasts will be copied
+  with src's labels.
 """
-function copy_forecasts!(src::InfrastructureSystemsType, dst::InfrastructureSystemsType)
+function copy_forecasts!(
+    src::InfrastructureSystemsType,
+    dst::InfrastructureSystemsType,
+    label_mapping::Union{Nothing, Dict{String, String}} = nothing,
+)
     for forecast in iterate_forecasts(ForecastInternal, src)
-        add_forecast!(dst, forecast)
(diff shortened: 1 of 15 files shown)


src/components.jl
```diff
@@ -215,7 +215,7 @@ function get_components_by_name(
     components::Components,
     name::AbstractString,
 )::Vector{T} where {T <: InfrastructureSystemsType}
-    if !isabstracttype(T)
+    if isconcretetype(T)
         throw(ArgumentError("get_components_by_name only supports abstract types: $T"))
```
daniel-thom, line 219: Can you change the error string?  ```does not support concrete types```

jd-lara:
bump @kdheepak for this review. I will leave it open for another 24 hours for comments.

src/component.jl
```diff
@@ -390,6 +390,47 @@ function check_contiguous_forecasts(
...
+
+# Arguments
+- `src::InfrastructureSystemsType`: Source component
+- `dst::InfrastructureSystemsType`: Destination component
+- `label_mapping::Dict = nothing`: Optionally map src labels to different dst labels.
+  If provided and src has a forecast with a label not present in label_mapping, that
+  forecast will not copied. If label_mapping is nothing then all forecasts will be copied
+  with src's labels.
+"""
+function copy_forecasts!(
```
daniel-thom, line 405: Why are these changes showing up?  They are in master.

src/utils/utils.jl
```diff
@@ -309,3 +309,89 @@ macro scoped_enum(T, args...)
...
+function forward(sender::Tuple{Type, Symbol}, receiver::Type)
+    code = Vector{String}()
+    active_methods = getfield.(InteractiveUtils.methodswith(sender[1]), :name)
+    for m in InteractiveUtils.methodswith(receiver)
+        m.name ∈ active_methods && continue
+        if startswith(string(m.name), "get_") && m.nargs == 2
+            # forwarding works for functions with 1 argument and starts with `get_`
+            append!(code, forward(sender, receiver, m))
+        end
+        if startswith(string(m.name), "set_") && m.nargs == 3
```
daniel-thom, line 381: elseif

src/component.jl
jd-lara: I don't know, I think that when I rebased they showed up here.

PR #99, opened 2020-06-08
PSIMA-62:Adding check limits for Int

sourabhdalvi:


src/validation.jl
```diff
@@ -166,6 +166,15 @@ function check_limits(
     return check_limits_impl(valid_info, field_value)
 end
 
+function check_limits(
```
jd-lara, line 169: Why this and not just changing the function in line 160 to take `Union{Nothing, Real}` ?

daniel-thom: Good point

sourabhdalvi: I agree, makes more sense to do that

PR #102, opened 2020-07-07
add contional on needs_conversion

jd-lara:


src/utils/generate_structs.jl
```diff
@@ -61,7 +61,7 @@ end
 {{/has_null_values}}
 {{#accessors}}
 {{#create_docstring}}\"\"\"Get {{struct_name}} {{name}}.\"\"\"{{/create_docstring}}
-{{accessor}}(value::{{struct_name}}) = value.{{name}}
+{{accessor}}(value::{{struct_name}}) = {{#needs_conversion}}get_value(value, :{{name}}){{/needs_conversion}}{{^needs_conversion}}value.{{name}}{{/needs_conversion}}
```
daniel-thom, line 64: What do you think about changing `needs_conversion` to `conversion_function` and then replace `get_value` with that?  Then the user could specify any custom function.

src/utils/generate_structs.jl
```diff
@@ -61,7 +61,7 @@ end
 {{/has_null_values}}
 {{#accessors}}
 {{#create_docstring}}\"\"\"Get {{struct_name}} {{name}}.\"\"\"{{/create_docstring}}
-{{accessor}}(value::{{struct_name}}) = value.{{name}}
+{{accessor}}(value::{{struct_name}}) = {{#needs_conversion}}get_value({{data_type}}, value, :{{name}}){{/needs_conversion}}{{^needs_conversion}}value.{{name}}{{/needs_conversion}}
```
daniel-thom, line 64: I suggest that this would be a better end result of generated code.
```
get_active_power(value::ThermalStandard)= get_value(value, value.active_power)
```
Dispatch would work on the type of the field and you wouldn't have to include it explicitly.

jd-lara: I feel is the same thing, but sure I can simplify it.

daniel-thom: It's very possible that I am misunderstanding the complete flow.  I was expecting to see something like this:

```
# Auto-generated
get_active_power(value::ThermalStandard) = convert_power(value, value.active_power)
```

```
function convert_power(c::Component, value::Float64)
    units_settings = get_units_settings(c)
    # do math on value
    return converted_value
end

function convert_power(c::Component, value::MinMax)
    units_settings = get_units_settings(c)
    # do math on value
    return converted_value
end
```

Can you point me to an example of what you were thinking?

src/internal.jl
```diff
@@ -55,3 +66,31 @@ Assign a new UUID.
...
+    for field in fields
+        val = getfield(internal, field)
+        if val isa UnitsData
+            val = nothing
+            push!(vals, val)
+            push!(final_fields, field)
+        else
+            push!(vals, val)
+            push!(final_fields, field)
+        end
```
daniel-thom, line 92:
```suggestion
        if val isa UnitsData
            val = nothing
        end
        push!(vals, val)
        push!(final_fields, field)
```

daniel-thom: Can you add a comment explaining why this is necessary?

src/utils/generate_structs.jl
daniel-thom: We should discuss this to reach a resolution.

jd-lara: What else is left to discuss?

PR #107, opened 2020-07-16
Remove container when the last component is removed

daniel-thom:
This was not really a bug, but it was causing serialization tests to fail
because empty containes were not getting deserialized.


jd-lara:
@daniel-thom can you confirm this is ready to merge?

PR #116, opened 2020-08-18
Add filtering to iterate_forecasts

daniel-thom:
This improves the interface to the `iterate_forecasts` function for both the system and a component.  You can filter on any parameter.

Note that this branch includes #115

### DIFF
--- a/src/component.jl
+++ b/src/component.jl
@@ -448,17 +448,49 @@ end
 
 """
 Returns an iterator of Forecast instances attached to the component.
+
+Note that passing a filter function can be much slower than the other filtering parameters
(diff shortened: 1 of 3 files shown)


src/component.jl
```diff
@@ -448,17 +448,49 @@ end
 
 """
 Returns an iterator of Forecast instances attached to the component.
```
jd-lara, line 450: Are you planning on importing these methods later in PSY?

daniel-thom: Yes.  For `System` we will have to change the existing function to be `IS.iterate_forecasts` and to forward the new parameters.

PR #117, opened 2020-08-19
Fix printing of undefined components

raphaelsaavedra:
Currently, creating a vector of undefined components results in an error:

```
julia> buses = Vector{Bus}(undef, 5)
5-element Array{Bus,1}
Error showing value of type Array{Bus,1}:
ERROR: UndefRefError: access to undefined reference
```

On the other hand, `buses = Vector{Bus}(undef, 5);` works, because the issue is related to `show`. This PR fixes this error:

```
julia> buses = Vector{Bus}(undef, 5)
5-element Array{Bus,1}
#undef
#undef
#undef
#undef
#undef
```

### DIFF
--- a/src/utils/print.jl
+++ b/src/utils/print.jl
@@ -131,7 +131,7 @@ end
 
 function Base.show(io::IO, ::MIME"text/plain", ists::Vector{<:InfrastructureSystemsType})
     println(io, summary(ists))
-    for i = 1:length(ists)
+    for i in 1:length(ists)
         if isassigned(ists, i)
             println(io, "$(summary(ists[i]))")
         else
--- a/src/utils/print.jl
+++ b/src/utils/print.jl
@@ -131,8 +131,8 @@ end
 
 function Base.show(io::IO, ::MIME"text/plain", ists::Vector{<:InfrastructureSystemsType})
     println(io, summary(ists))
-    for ist in ists
-        println(io, "$(summary(ist))")
+    for i = 1:length(ists)
+        isassigned(ists, i) && println(io, "$(summary(ists[i]))")
     end
 end
 
--- a/src/utils/print.jl
+++ b/src/utils/print.jl
@@ -132,7 +132,11 @@ end
 function Base.show(io::IO, ::MIME"text/plain", ists::Vector{<:InfrastructureSystemsType})
     println(io, summary(ists))
     for i = 1:length(ists)
-        isassigned(ists, i) && println(io, "$(summary(ists[i]))")
+        if isassigned(ists, i)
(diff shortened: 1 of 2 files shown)


src/utils/print.jl
```diff
@@ -131,8 +131,8 @@ end
 
 function Base.show(io::IO, ::MIME"text/plain", ists::Vector{<:InfrastructureSystemsType})
     println(io, summary(ists))
-    for ist in ists
-        println(io, "$(summary(ist))")
+    for i = 1:length(ists)
+        isassigned(ists, i) && println(io, "$(summary(ists[i]))")
```
nickrobinson251, line 135: i think we still want to print the undefined values but to print them as `#undef`, i.e. printing a length 3 Vector should show 3 things, it's just that we want to avoid trying to access undefined references. 

e.g.
```julia
     for i = 1:length(ists) 
         if isassigned(ists, i)
             println(io, summary(ists[i]))
         else
            println(io, Base.undef_ref_str)
        end
```

nickrobinson251 (commented):
:+1:   Nice little user-experience fix

src/utils/print.jl
raphaelsaavedra: Good point, I'll add that. By the way, do you know why format check is failing? I'm not familiar with that.

nickrobinson251:
it'd also be good to add a simple test, taking inspiration from `test.show.jl` in Base
https://github.com/JuliaLang/julia/blob/032cbc20f4ba53d4b4d3c3d13f8a1915d9cbb67e/test/show.jl#L25
e.g. a test using `Vector{TestComponent}(undef, 3)`

raphaelsaavedra:
Sure. What would be a good test? Is something like this enough?

```
@testset "Test component prints" begin
    v = Vector{PowerSystemType}(undef, 3)
    @test println(v) === nothing
end
```

nickrobinson251:
yeah, `sprint(show, v) == "..."` is the way to go, i think

nickrobinson251 (approved):
this LGTM!

andrewrosemberg:
Need to call `include(".github/workflows/formatter_code.jl")` in the rood directory to format the code (using JuliaFormater).

test/test_utils.jl
```diff
@@ -89,3 +89,9 @@ end
     @test Fruits.APPLE isa Fruits.Fruit
     @test Fruits.ORANGE isa Fruits.Fruit
 end
+
+@testset "Test undef component prints" begin
```
andrewrosemberg, line 93: maybe do a test with both #undef and defined value to increase coverage? not sure if it's needed

PR #120, opened 2020-08-31
Refactor serialization

daniel-thom:
Refactor serialization to use JSON3
    
- Replaced JSON2 with JSON3.
- Serialize to standard types with custom methods and then to JSON.
- Added kwarg-only constructors to auto-generated structs.
- Remove serialization support from SystemData so that parent containers
can control the serialization/deserialization process.
- Added InfrastructureSystemsComponent to differentiate component
structs from generic IS structs.


src/descriptors/structs.json
```diff
@@ -1,244 +1,247 @@
...
-        "name": "internal",
-        "data_type": "InfrastructureSystemsInternal",
-        "internal_default": "InfrastructureSystemsInternal()"
-      }
-    ],
-    "supertype": "ForecastInternal"
-  }
-]
+{
+  "auto_generated_structs": [
```
jd-lara, line 2: Are the changes here the addition of a second level? The diff is the whole file.

src/time_series_data.jl
```diff
@@ -1,4 +1,4 @@
-struct TimeSeriesData <: InfrastructureSystemsType
+struct TimeSeriesData <: InfrastructureSystemsComponent
```
jd-lara, line 1: Add Docstring here

src/InfrastructureSystems.jl
```diff
@@ -20,24 +19,43 @@ using DocStringExtensions
...
+All structs must implement a kwarg-only constructor to allow deserializing from a Dict.
+"""
 abstract type InfrastructureSystemsType end
 
+"""
+Base type for structs that are stored in a system.
+
+Required interface functions for subtypes:
+- get_name()
+- get_internal()
```
jd-lara, line 33: Should we add a note about the cases where a component has a forecast?

src/descriptors/structs.json
daniel-thom: The change is actually very small.  I moved the struct descriptors into a new JSON object called `auto_generated_structs` in order to match the format we use in PSY.  The contents did not change.

PR #122, opened 2020-09-14
Serialize component modules

daniel-thom:
This PR adds support for serializing types that reside outside of PowerSystems.  The serialization process adds a metadata dict to each component dict that includes module and type information.


src/serialization.jl
```diff
@@ -58,7 +63,49 @@ function serialize(vals::Vector{T}) where {T <: InfrastructureSystemsType}
...
+Return the type information for the serialized struct.
+"""
+get_serialization_metadata(data::Dict) = data[METADATA_KEY]
+
+function get_type_from_serialization_data(data::Dict)
+    return get_type_from_serialization_metadata(get_serialization_metadata(data))
+end
+
+function get_type_from_serialization_metadata(metadata::Dict)
+    mod = Base.root_module(Base.__toplevel__, Symbol(metadata[MODULE_KEY]))
```
daniel-thom, line 97: @kdheepak FYI, in case you're interested.  I implemented the changes we discussed here.  Also note the use of `root_module` here.  It allows us to find modules above the current scope.

claytonpbarrows (approved):

PR #123, opened 2020-09-14
Rename forecasts to time series

daniel-thom:
This PR primarily renames `forecast` to `time_series`.  It also does the following:
- Renames `iterate_forecasts` to `get_time_series_multiple`.
- Extends the former `get_forecast_values` into `get_time_series_arrays`, `get_time_series_timestamps`, and `get_time_series_values`.  This allows users to get the underlying TimeArray, timestamps, or values.
- Renames `scaling_factor` to `normalization_factor`.
- Adds `scaling_factor_multiplier` to all time series.

This requires the PowerSystems branch of the same name for cross-project testing to pass.


src/generated/PiecewiseFunction.jl
```diff
@@ -2,31 +2,35 @@
 This file is auto-generated. Do not edit.
 =#
 """
-    mutable struct PiecewiseFunction <: Forecast
+    mutable struct PiecewiseFunction <: TimeSeriesData
```
jd-lara, line 5: When did we implement this struct? It seems redundant to what we are planning on doing for the Variable Cost.

src/generated/ScenarioBased.jl
```diff
@@ -2,31 +2,35 @@
 This file is auto-generated. Do not edit.
 =#
 """
-    mutable struct ScenarioBased <: Forecast
+    mutable struct ScenarioBased <: TimeSeriesData
```
jd-lara, line 5: We can rename ScenatioBased to just Scenarios. @claytonpbarrows what do you think? I find awkward `ScenarioBased` time series.

src/system_data.jl
```diff
@@ -51,294 +51,319 @@ function SystemData(;
...
     )
 end
 
 """
-Adds forecasts from a metadata file or metadata descriptors.
+Adds time_series from a metadata file or metadata descriptors.
 
 # Arguments
-- `::Type{T}`: forecasted component type; may be abstract
+- `::Type{T}`: time_seriesed component type; may be abstract
```
jd-lara, line 81: typo?

jd-lara:
The PR looks great in terms of new capabilities and implementation. We will probably need to test drive it more. I have a comment - question about the PWL forecast object. We should define if we want to keep it in 1.0 or not.

src/generated/ScenarioBased.jl
claytonpbarrows: I also find it to be awkward. But I don't love `Scenarios` either. Also, with `TimeSeriesData` do we even need `ScenarioBased` now that we have both `label` and `accessor`? If the answer is yes, then I'm fine with the proposal.

src/generated/PiecewiseFunction.jl
daniel-thom: https://github.com/NREL-SIIP/InfrastructureSystems.jl/pull/90

daniel-thom: Removed.

jd-lara (approved):
LGTM

PR #124, opened 2020-09-20
use run instead of cp

jd-lara:


src/hdf5_time_series_storage.jl
```diff
@@ -76,7 +76,7 @@ function from_file(
     else
         file_path, io = mktemp()
         close(io)
-        cp(filename, file_path; force = true)
+        run(`cp -f $(filename) $(file_path)`)
```
daniel-thom, line 79: This should fail on some Windows systems.  If users have installed GNU coreutils then it will work.  I think it would be safer to check the platform.  I'm not sure what is installed on the CI systems.

```
if Sys.iswindows()
    run(`copy /Y $filename $file_path`)
else
    run(`cp -f $(filename) $(file_path)`)
end
```
It would also be better to put this in a function.  And add a comment describing the Julia bug (with a link) so that future devs can check on removing it.

jd-lara: The proposed substitution failed in windows.

jd-lara: The final solution was to use a try/catch with a call to cmd before the copy command. I tested the command ``run(`cmd \c copy $src $dst)`)`` in the previous CI run and it works

src/utils/utils.jl
```diff
@@ -374,3 +374,18 @@ macro forward(sender, receiver, exclusions = Symbol[])
...
+# This function is used instead of cp given
+# https://github.com/JuliaLang/julia/issues/30723
+function copy_file(src::AbstractString, dst::AbstractString)
+    try
+        run(`cp -f $(src) $(dst)`)
+    catch e
+        if Sys.iswindows()
+            run(`cmd /c copy /Y $(src) $(dst)`)
+        else
+            return e
```
daniel-thom, line 387: I think you want `rethrow` instead

src/utils/utils.jl
```diff
@@ -374,3 +374,18 @@ macro forward(sender, receiver, exclusions = Symbol[])
...
 end
+
+# This function is used instead of cp given
+# https://github.com/JuliaLang/julia/issues/30723
+function copy_file(src::AbstractString, dst::AbstractString)
+    try
+        run(`cp -f $(src) $(dst)`)
+    catch e
+        if Sys.iswindows()
+            run(`cmd /c copy /Y $(src) $(dst)`)
```
daniel-thom, line 385: Sorry for the bad suggestion. Don't know why `cmd /c` is required.

jd-lara: it seems that it depends on the shell interpreter on the windows machine.

src/utils/utils.jl
```diff
@@ -384,7 +384,7 @@ function copy_file(src::AbstractString, dst::AbstractString)
         if Sys.iswindows()
             run(`cmd /c copy /Y $(src) $(dst)`)
         else
-            return e
+            rethrow(e)
```
daniel-thom, line 387: This is technically correct but not the recommended way.  Just `rethrow()`.
```
  rethrow()

  Rethrow the current exception from within a catch block. The rethrown exception will continue propagation as if it
  had not been caught.

  │ Note
  │
  │  The alternative form rethrow(e) allows you to associate an alternative exception object e with the current
  │  backtrace. However this misrepresents the program state at the time of the error so you're encouraged to
  │  instead throw a new exception using throw(e). In Julia 1.1 and above, using throw(e) will preserve the
  │  root cause exception on the stack, as described in catch_stack.
```

daniel-thom (approved):
Approved but you should remove the unnecessary `e`.

src/utils/utils.jl
jd-lara: removed via 4db7838

PR #125, opened 2020-09-21
Add Deterministic constructors to support all input formats

daniel-thom:
1. Allow user to construct `Deterministic` time series data from a CSV file or DataFrame, and support a normalization factor.
2. Removes `add_time_series!` methods that constructed `Deterministic` and added it to the system in one step. The user will need to perform these steps independently.
3. Adds `add_time_series!` method that supports multiple components.


jd-lara (approved):
LGTM, but it needs re running the formatter code

PR #126, opened 2020-09-24
time series split part 1

jd-lara:
Part 1 of the upgrades.


src/time_series_container.jl
```diff
@@ -73,72 +71,69 @@ end
...
-function get_time_series_initial_times(container::TimeSeriesContainer)
-    initial_times = Set{Dates.DateTime}()
-    for key in keys(container.data)
-        push!(initial_times, key.initial_time)
-    end
+function get_time_series_initial_times(ts_metadata::TimeSeriesMetadata)
+    initial_time_stamp = get_initial_time_stamp(ts_metadata)
+    interval = get_interval(ts_metadata)
+    count = get_count(ts_metadata)
+    return collect(range(initial_time_stamp; length = count, step = interval))
```
daniel-thom, line 88: Does this need to be `collect`ed?

daniel-thom: I think this should live in the same file as TimeSeriesMetadata.

src/time_series_container.jl
```diff
@@ -73,72 +71,69 @@ end
...
+end
 
-    return sort!(Vector{Dates.DateTime}(collect(initial_times)))
+function get_time_series_initial_times(container::TimeSeriesContainer)
+    ts_metadata = first(values(container.data))
+    return get_time_series_initial_times(ts_metadata)
```
daniel-thom, line 93: I'm not sure this method makes sense.  The user could store static time series and forecasts and so the answer may not be true for all.  If the users stores one forecast with labelX and another with labelY, we don't technically need to impose a requirement that those forecasts have the same initial times.  I'm not saying that that isn't a good restriction; just pointing out that each forecast is now completely independent.

src/time_series_parameters.jl
```diff
@@ -50,6 +51,31 @@ function check_add_time_series!(params::TimeSeriesParameters, ts::TimeSeriesMeta
...
+
+function _verify_time_series(
+    params::TimeSeriesParameters,
+    time_series::SingleTimeSeriesMetadata,
+)
+    if time_series.resolution != params.resolution
+        throw(DataFormatError(
+            "time series resolution $(time_series.resolution) does not match system " *
+            "resolution $(params.resolution)",
+        ))
```
daniel-thom, line 65: Per the new forecast independence, I'm not sure we need this requirement.

src/hdf5_time_series_storage.jl
```diff
@@ -85,11 +85,22 @@ end
 
 get_file_path(storage::Hdf5TimeSeriesStorage) = storage.file_path
 
+function _time_array_wrapper_to_array(ta::TimeDataContainer)
+    # TODO: Implement for more dimensions.
+    # TODO: Is this storing the data efficiently?
+    if length(ta.data) == 1
+        return TimeSeries.values(first(values(ta.data)))
+    else
+        length(ta.data) > 1
```
daniel-thom, line 94: delete

src/hdf5_time_series_storage.jl
```diff
@@ -85,11 +85,22 @@ end
 
 get_file_path(storage::Hdf5TimeSeriesStorage) = storage.file_path
 
+function _time_array_wrapper_to_array(ta::TimeDataContainer)
```
daniel-thom, line 88: This should live in the file containing TimeDataContainer.  And it needs a new name.

src/hdf5_time_series_storage.jl
```diff
@@ -85,11 +85,22 @@ end
...
+        return hcat(TimeSeries.values.(values(ta.data))...)
+    end
+end
+
 function add_time_series!(
     storage::Hdf5TimeSeriesStorage,
     component_uuid::UUIDs.UUID,
     label::AbstractString,
-    ta::TimeArrayWrapper,
+    ta::TimeDataContainer,
```
daniel-thom, line 103: Need a new variable name.

src/hdf5_time_series_storage.jl
```diff
@@ -182,43 +200,78 @@ function remove_time_series!(
     end
 end
 
+function _make_time_array(timestamps, data, columns::Union{Vector{Symbol}, Nothing})
```
daniel-thom, line 203: For an internal function like this I suggest leaving `columns` untyped.

src/hdf5_time_series_storage.jl
```diff
@@ -182,43 +200,78 @@ function remove_time_series!(
...
     storage::Hdf5TimeSeriesStorage,
-    uuid::UUIDs.UUID;
-    index = 0,
-    len = 0,
-)::TimeSeries.TimeArray
+    uuid::UUIDs.UUID,
+    index::Int,
+    len::Int,
+    count::Int,
+)::DataStructures.SortedDict{Dates.DateTime, TimeSeries.TimeArray}
```
daniel-thom, line 217: The return type shouldn't have been there before. You can remove.

src/hdf5_time_series_storage.jl
```diff
@@ -182,43 +200,78 @@ function remove_time_series!(
...
-        # If the user set column names, return them. Otherwise, let TimeArray pick them.
-        dtimestamps = [Dates.epochms2datetime(x) for x in timestamps]
-        if HDF5.exists(HDF5.attrs(path), "columns")
-            columns = Symbol.(HDF5.read(HDF5.attrs(path)["columns"]))
-            return TimeSeries.TimeArray(dtimestamps, data, columns)
+            #Making a Dict prevents type instability in the return of the function
+            return DataStructures.SortedDict(
+                time_stamps[1] => _make_time_array(time_stamps, data, columns),
+            )
+        elseif HDF5.exists(HDF5.attrs(path), "interval")
```
daniel-thom, line 254:
```suggestion
        else
```

src/hdf5_time_series_storage.jl
```diff
@@ -182,43 +200,78 @@ function remove_time_series!(
...
         root = _get_root(storage, file)
         path = _get_time_series_path(root, uuid)
-
-        if index == 0
-            data = HDF5.read(path["data"])
-            timestamps = HDF5.read(path["timestamps"])
+        _initial_time_stamp = HDF5.read(HDF5.attrs(path)["initial_time"])
+        initial_time_stamp = Dates.epochms2datetime(_initial_time_stamp)
+        resolution = Dates.Millisecond(HDF5.read(HDF5.attrs(path)["resolution"]))
+        series_length = HDF5.read(HDF5.attrs(path)["length"])
```
daniel-thom, line 224: This information is already provided by HDF and so we don't need to store it.  It can be retrieved with `size(path["data"])`.

src/hdf5_time_series_storage.jl
```diff
@@ -182,43 +200,78 @@ function remove_time_series!(
...
+        # TODO: Consider case Interval_ms = 0 if contigous instead of not existing
+        if !HDF5.exists(HDF5.attrs(path), "interval")
+            @assert HDF5.read(HDF5.attrs(path)["count"]) == 1
+            @debug "reconstructing a contigouos time series"
+            time_stamps =
+                range(initial_time_stamp; length = series_length, step = resolution)
+            if len == series_length && index == 1
+                data = HDF5.read(path["data"])
+            elseif len <= series_length
+                end_index = min(index + len - 1, series_length)
```
daniel-thom, line 244: 1. This allows the user to pass an invalid valid for `len`.  That will just lead to more problems for them.  We should throw an error in that case.
2. I suspect that this is an off-by-one error.  `index + len`

src/hdf5_time_series_storage.jl
```diff
@@ -182,43 +200,78 @@ function remove_time_series!(
...
 function get_time_series(
     storage::Hdf5TimeSeriesStorage,
-    uuid::UUIDs.UUID;
-    index = 0,
-    len = 0,
-)::TimeSeries.TimeArray
+    uuid::UUIDs.UUID,
+    index::Int,
+    len::Int,
+    count::Int,
```
daniel-thom, line 216: I think these should have default values, as before.  Return all data.  It reduces the burden on higher-level code.

src/generated/SingleTimeSeriesMetadata.jl
```diff
@@ -0,0 +1,79 @@
...
+- `resolution::Dates.Period`
+- `initial_time::Dates.DateTime`: time series availability time
+- `time_series_uuid::UUIDs.UUID`: reference to time series data
+- `length::Int`: length of this time series
+- `scaling_factor_multiplier::Union{Nothing, Function}`: Applicable when the time series data are scaling factors. Called on the associated component to convert the values.
+- `internal::InfrastructureSystemsInternal`
+"""
+mutable struct SingleTimeSeriesMetadata <: StaticTimeSeriesMetadata
+    "user-defined label"
+    label::String
```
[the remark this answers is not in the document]
jd-lara: Decided to change later.

src/hdf5_time_series_storage.jl
jd-lara: without this, a single time series get stored as a row and not as a column.

src/hdf5_time_series_storage.jl
jd-lara: ok, I will move it

src/hdf5_time_series_storage.jl
daniel-thom: This line is a no-op.  The expression returns a boolean but it is not captured.

src/time_array_container.jl
```diff
@@ -0,0 +1,67 @@
...
+    data::DataStructures.SortedDict{Dates.DateTime, Vector{Float64}}
+    resolution::Dates.Period
+    internal::InfrastructureSystemsInternal
+
+    function TimeDataContainer(data, resolution, internal)
+        for v in values(data)
+            if length(v) < 2
+                throw(ArgumentError("time array length must be at least 2"))
+            end
+        end
```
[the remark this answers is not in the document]
jd-lara: Yes, will add it

src/hdf5_time_series_storage.jl
jd-lara: ta -> tc ?

src/hdf5_time_series_storage.jl
jd-lara: Don't we want to catch inadequate file structures?

src/hdf5_time_series_storage.jl
jd-lara: Do we need to keep this functionality? It seems this is an out dated thing

src/component.jl
```diff
@@ -36,61 +36,89 @@ function clear_time_series!(component::InfrastructureSystemsComponent)
...
 """
 function get_time_series(
     ::Type{T},
     component::InfrastructureSystemsComponent,
     initial_time::Dates.DateTime,
-    label::AbstractString,
-    horizon::Int,
+    label::AbstractString;
+    horizon::Union{Nothing, Int} = nothing,
+    count::Int = 1,
```
daniel-thom, line 93: Should T be defined as Forecast?

src/hdf5_time_series_storage.jl
daniel-thom: tc, container, ...

src/hdf5_time_series_storage.jl
daniel-thom: If the user sets the column name in the time array then this allows that preference to be preserved.

src/hdf5_time_series_storage.jl
daniel-thom: You don't need to re-read interval. The if statement is sufficient to make this just `else`.

src/hdf5_time_series_storage.jl
daniel-thom: Actually, TimeDataContainer now only stores initial times and arrays of data. We will have to document that column names are not preserved in DataFrames and TimeArrays.

src/hdf5_time_series_storage.jl
jd-lara: move to size. Leave comments relating to the meta data names

src/hdf5_time_series_storage.jl
jd-lara: Checked, no off-by-one error. Agreed on the improvement of conditional

daniel-thom (approved):
This gets a base level working.  I'll branch off this to enable the interfaces for PowerSystems.  And fix the in-memory time series storage.

PR #127, opened 2020-09-25
Fix row and column time series indices

daniel-thom:


test/test_time_series.jl
```diff
@@ -575,405 +582,6 @@ end
     )
 end
 
-function validate_generated_initial_times(
```
jd-lara, line 578: Do we still want to have some of the functionalities from these tests?

jd-lara (approved):

test/test_time_series.jl
daniel-thom: I'm pretty sure this is all pointless now

PR #128, opened 2020-09-25
add csv forecast parser part 1

jd-lara:
adds the parser for forecasts, supporting single DateTime in the first column format.


src/time_series_formats.jl
```diff
@@ -25,6 +25,33 @@ function read_time_series(file_path::AbstractString, component_name = nothing; k
     return read_time_series(get_time_series_format(file), file, component_name; kwargs...)
 end
 
+"""
+Return a TimeDataContaier from a CSV file.
```
daniel-thom, line 29:
```suggestion
Return a TimeDataContainer from a CSV file.
```

src/time_series_formats.jl
```diff
@@ -25,6 +25,33 @@ function read_time_series(file_path::AbstractString, component_name = nothing; k
...
+
+Pass component_name when the file does not have the component name in a column header.
```
daniel-thom, line 31: How do you envision attaching this to a specific component?

src/time_series_formats.jl
```diff
@@ -25,6 +25,33 @@ function read_time_series(file_path::AbstractString, component_name = nothing; k
...
+"""
+function read_forecast_from_CSV(file_path::AbstractString, resolution::Dates.Period)
```
daniel-thom, line 33:
```suggestion
function read_forecast_from_csv(file_path::AbstractString, resolution::Dates.Period)
```

src/time_series_formats.jl
```diff
@@ -25,6 +25,33 @@ function read_time_series(file_path::AbstractString, component_name = nothing; k
     return read_time_series(get_time_series_format(file), file, component_name; kwargs...)
 end
 
+"""
```
daniel-thom, line 28: Should we make our own format in JSON?  We could include other information like resolution and component name/type.  Or are you thinking that a separate `timeseries_pointers.json` would have that information and point to this file?

src/time_series_formats.jl
jd-lara: Can we use a similar pointers file but instead of going from columns -> component it will be file -> component?

src/time_series_formats.jl
jd-lara: I think using the same pointers files with extension is the way to go.

daniel-thom: My preference is one JSON file that contains all relevant data and metadata.  It would be very clear to users on what they need to provide.  I think the reason that we have the timeseries_pointers.json combined with raw CSV is that we are working with existing file formats.  If the same thing is true for this type of forecast data then I'm certainly OK staying with it.  However, if the user is going to have to modify a bunch of things anyway, I would prefer to go with one self-contained file that is extensible.

PR #130, opened 2020-09-25
Refactors the workflow of adding and getting time series

daniel-thom:


src/time_array_container.jl
```diff
@@ -1,82 +0,0 @@
...
-    resolution::Dates.Period
-    internal::InfrastructureSystemsInternal
-
-    function TimeDataContainer(
-        data::SortedDict{Dates.DateTime, Array{T, N}},
-        resolution::Dates.Period,
-        internal::InfrastructureSystemsInternal,
-    ) where {T, N}
-        series_length = length(first(values(data)))
-        for (k, v) in data
```
jd-lara, line 12: I understand the removal of TimeDataContainer but couldn't find where did these checks go.

jd-lara (approved):
LGTM, we should consider the checks on the array sizes.

src/descriptors/structs.json
```diff
@@ -54,46 +54,65 @@
...
+        },
+        {
+          "name": "resolution",
+          "data_type": "Dates.Period",
+          "comment": "forecast resolution"
+        },
         {
           "name": "data",
-          "data_type": "SortedDict{Dates.DateTime, TimeSeries.TimeArray}",
+          "data_type": "SortedDict{Dates.DateTime, Array}",
```
jd-lara, line 81: Vector in the case of Deterministic?

src/time_array_container.jl
daniel-thom: Good catch.  fixed.

src/descriptors/structs.json
daniel-thom: Fixed

PR #132, opened 2020-09-25
Fix parsing of time series from file metadata

daniel-thom:


jd-lara (approved):
LGTM, do we want to add code similar to the resolution check for intervals?

PR #133, opened 2020-09-26
Remove initial time from interface functions

daniel-thom:


src/system_data.jl
```diff
@@ -200,14 +200,13 @@ function remove_time_series!(
...
     data::SystemData,
     component::InfrastructureSystemsComponent,
-    initial_time::Dates.DateTime,
     name::String,
 ) where {T <: TimeSeriesData}
     type = time_series_data_to_metadata(T)
-    time_series = get_time_series(type, component, initial_time, name)
+    time_series = get_time_series(type, component, name)
     uuid = get_time_series_uuid(time_series)
     # TODO: can this be atomic?
```
jd-lara, line 208: Is the comment still valid?

src/system_data.jl
```diff
@@ -332,25 +332,18 @@ Call `collect` on the result to get an array.
 - `data::SystemData`: system
 - `filter_func = nothing`: Only return time_series for which this returns true.
 - `type = nothing`: Only return time_series with this type.
-- `initial_time = nothing`: Only return time_series matching this value.
 - `name = nothing`: Only return time_series matching this value.
 """
 function get_time_series_multiple(
```
jd-lara, line 337: I don't know if we need to keep this function. Will just return all the `TimeSeriesData` objects or the data stored in the forecasts?

jd-lara (approved):
LGTM, I am not clear still on what is the usage for get_time_series multiple

daniel-thom:
> 
> 
> LGTM, I am not clear still on what is the usage for get_time_series multiple

I think it is a very helpful debug/interactive function, especially when called on the system.

src/system_data.jl
daniel-thom: It's hopeful but can be removed if I can't think of anything before we merge to master.

src/system_data.jl
daniel-thom: It returns all TimeSeriesData objects

PR #134, opened 2020-09-28
Time series updates

daniel-thom:


src/deterministic.jl
```diff
@@ -0,0 +1,130 @@
...
+                throw(ArgumentError("The values in the data dict can't be converted to TimeArrays. Resulting error: $e"))
+            end
+        end
+    end
+
+    ta = handle_normalization_factor(ta, normalization_factor)
+    return Deterministic(name, ta, scaling_factor_multiplier)
+end
+
+# TODO: need to make concatenation constructors for Probabilistic
```
jd-lara, line 43: Is it valid? yes -> Move. No -> Remove ?

src/deterministic.jl
```diff
@@ -0,0 +1,130 @@
...
+end
+
+function get_array_for_hdf(forecast::Deterministic)
+    return hcat(values(forecast.data)...)
+end
+
+"""
+Creates a new Deterministic from an existing instance and a subset of data.
+"""
+function split_time_series(
```
jd-lara, line 111: Should this be T <: StaticTimeSeries ?

src/in_memory_time_series_storage.jl
```diff
@@ -21,7 +21,7 @@ end
 
 function InMemoryTimeSeriesStorage()
     storage = InMemoryTimeSeriesStorage(Dict{UUIDs.UUID, _TimeSeriesRecord}())
-    @info "Created InMemoryTimeSeriesStorage"
+    @debug "Created InMemoryTimeSeriesStorage"
```
jd-lara, line 24: Should we delete this completely? Seems that it can create a huge amount of entries into the logs.

src/probabilistic.jl
```diff
@@ -0,0 +1,72 @@
...
+        ones(time_steps, length(percentiles)),
+    )
+
+    return Probabilistic(; name = name, percentiles = percentiles, data = data)
+end
+
+"""
+Constructs Probabilistic forecast after constructing a TimeArray from initial_time and time_steps.
+"""
+# TODO: do we need this check still?
```
jd-lara, line 22: The answer is yes, this is an important check, we could make it later

jd-lara (approved):
Looks like a great set of changes, comments to be resolved later.

src/deterministic.jl
daniel-thom: No, it is correct.  This function creates a new deterministic where all the fields except `data` and `internal` are the same as the original.  It would probably be more clear as a constructor.  I'll make the change.

src/in_memory_time_series_storage.jl
daniel-thom: It is only generated when the System is created.

src/deterministic.jl
daniel-thom: This was a long-standing TODO.

PR #135, opened 2020-09-28
Jd/add metadata forecast parse

jd-lara:
I went with a parametrized method to return the ts object in `make_time_series`


src/time_series_formats.jl
```diff
@@ -13,16 +12,24 @@ Return a TimeArray from a CSV file.
 
 Pass component_name when the file does not have the component name in a column header.
 """
-function read_time_series(file_path::AbstractString, component_name = nothing; kwargs...)
-    if !isfile(file_path)
-        msg = "TimeSeries file doesn't exist : $file_path"
+function read_time_series(metadata::TimeSeriesFileMetadata; kwargs...)
```
daniel-thom, line 15: It appears that the function `get_step_time` is no longer used.  Can all its methods and references in comments be deleted?

src/time_series_parser.jl
```diff
@@ -159,12 +171,15 @@ function handle_normalization_factor(
     return data
 end
 
+get_max_value(ta::TimeSeries.TimeArray) = maximum(TimeSeries.values(ta))
+get_max_value(ta::Vector) = maximum(ta)
+
 function handle_normalization_factor(
-    ta::TimeSeries.TimeArray,
+    ta::Union{TimeSeries.AbstractTimeSeries, AbstractArray},
```
daniel-thom, line 178: I think that we should implement this function on a TimeSeriesData type.  The original function returns a new TimeArray because TimeArrays are immutable.  In the case of Deterministic we can modify the data in place.

test/test_time_series.jl
```diff
@@ -150,584 +150,20 @@ end
...
     name = "Component1"
     component = IS.TestComponent(name, 5)
     IS.add_component!(data, component)
     @test !IS.has_time_series(component)
-
     file = joinpath(FORECASTS_DIR, "ComponentsAsColumnsNoTime.json")
     IS.add_time_series_from_file_metadata!(data, IS.InfrastructureSystemsComponent, file)
     @test IS.has_time_series(component)
 
-    all_time_series = get_all_time_series(data)
```
daniel-thom, line 163: Why did you delete all of these tests?

daniel-thom (changes requested):

src/time_series_parser.jl
```diff
@@ -16,9 +23,10 @@ mutable struct TimeSeriesFileMetadata
     normalization_factor::Union{String, Float64}
     "Path to the time series data file"
     data_file::String
+    "Resolution of the data being parsed in milliseconds"
+    resolution::Dates.Period
```
daniel-thom, line 27: @claytonpbarrows Note that this imposes a new requirement on the timeseries_pointers files.  They have to define `resolution`.  Previously the parsing code calculated it automatically.

test/data/time_series/DateTimeAsColumn_forecast.csv
```diff
@@ -0,0 +1,169 @@
+DateTime,1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19,20,21,22,23,24
```
daniel-thom, line 1: Should the file be called `DateTimeAsColumnDeterministic.csv`?

src/time_series_formats.jl
```diff
@@ -152,33 +159,44 @@ function get_timestamp(
...
+Pass component_name when the file does not have the component name in a column header.
 """
 function read_time_series(
     ::Type{T},
+    ::Type{U},
+    file::CSV.File;
+    kwargs...,
+) where {T <: TimeSeriesFileFormat, U <: Forecast}
+    error("The file format provided can't be parsed into a $U forecast")
+end
```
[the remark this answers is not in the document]
jd-lara: OK, I will remove it

src/deterministic.jl
```diff
@@ -128,3 +126,14 @@ function split_time_series(
 
     return T(vals...)
 end
+
+function make_time_series(info::TimeSeriesParsedInfo{Deterministic})
```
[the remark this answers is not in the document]
jd-lara: I agree that is more similar, but it was weird to me to have this pattern: 

```Julia
struct TimeSeriesInfo
    ts_type::DataType
    other_data::Dict
end

function make_time_series(val::TimeSeriesInfo
    return val.ts_type(val.other)
end
```

I think this is very strange. We call make to return a type we evaluate inside of another one.

src/time_series_parser.jl
jd-lara: Should we handle this in a later PR? It seems like an effort on its own.

src/deterministic.jl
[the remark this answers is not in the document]
daniel-thom: Here is an idea.  The time series type doesn't need to be part of TimeSeriesParseInfo because you added it to TimeSeriesFileMetadata.  You could change this function in system_data.jl line 225 to

```
function make_time_series!(cache::TimeSeriesCache, ts_file_metadata::TimeSeriesFileMetadata)
    info = add_time_series_info!(cache, ts_file_metadata)
    return ts_file_metadata.time_series_type(info)
end

```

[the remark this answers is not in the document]
jd-lara: I don't see how this is beneficial if we still need to implement the constructors based on info. I think the parametric version is easier to interpret.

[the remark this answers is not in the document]
daniel-thom: 1. The constructors in single_time_series.jsl and deterministic.jl will follow the same pattern.  Currently, you have the function make_time_series that behaves like a constructor but isn't named like one.
2. The type is already part of TimeSeriesFileMetadata, to which you already have access.  There isn't any value in duplicating it in TimeSeriesParsedInfo.
3. IMHO this isn't how people typically use parametric types.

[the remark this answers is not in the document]
daniel-thom: I think that having a function like `make_time_series` would be appropriate if it operated on an abstract type, obviating the need to add a constructor for each concrete type.  But that isn't what you've done.

src/time_series_parser.jl
daniel-thom: Agreed. Can you make a case for tracking?

src/time_series_parser.jl
```diff
@@ -254,35 +267,36 @@ function TimeSeriesParsedInfo(metadata::TimeSeriesFileMetadata, ta::TimeSeries.T
         metadata.component,
         metadata.name,
         normalization_factor,
-        ta,
+        raw_data,
         metadata.percentiles,
         metadata.data_file,
-        ts_type,
+        metadata.resolution,
+        metadata.time_series_type,
```
daniel-thom, line 274: Can this be removed?

jd-lara: `metadata.time_series_type` is used as the constructor for the time_series_data

daniel-thom: You're passing it to the constructor for TimeSeriesParserInfo. It is unused there.  You should be able to delete this line as well as 217.

src/deterministic.jl
```diff
@@ -40,6 +40,32 @@ function Deterministic(
     return Deterministic(name, ta, scaling_factor_multiplier)
 end
 
+"""
+Construct Deterministic from a CSV file. The file must have a column that is the name of the
```
daniel-thom, line 44: The docstring is incorrect.  Can you change it to a brief description of the format?

src/deterministic.jl
```diff
@@ -18,26 +18,64 @@ function Deterministic(
...
 )
     for (k, v) in data
         if v isa DataFrames.DataFrame
-            data[k] = TimeSeries.TimeArray(v; timestamp = timestamp)
+            data[k] = Array(v)
         elseif v isa TimeSeries.TimeArray
-            continue
+            data[k] = TimeSeries.values(v)
+            if resolution === nothing
+                resolution = diff(TimeSeries.timestamp(v))[1]
```
daniel-thom, line 29: This computes a diff of the entire timestamp array and then takes the first element.  You just need to subtract the first two elements.

src/deterministic.jl
```diff
@@ -18,26 +18,64 @@ function Deterministic(
...
-                    TimeSeries.TimeArray(range(k, length = length(v), step = resolution))
+                data[k] = Vector{Float64}()
             catch e
-                throw(ArgumentError("The values in the data dict can't be converted to TimeArrays. Resulting error: $e"))
+                throw(ArgumentError("The values in the data dict are in an invalid format. Resulting error: $e"))
             end
         end
     end
+    if resolution === nothing
+        throw(ArgumentError("Need to pass resolution keyword argument or the data most be provided using TimeArrays"))
```
daniel-thom, line 40:
```suggestion
        throw(ArgumentError("Need to pass resolution keyword argument or the data must be provided using TimeArrays"))
```

src/deterministic.jl
```diff
@@ -18,26 +18,64 @@ function Deterministic(
     data::Union{Dict{Dates.DateTime, Any}, SortedDict{Dates.DateTime, Any}};
     normalization_factor::NormalizationFactor = 1.0,
     scaling_factor_multiplier::Union{Nothing, Function} = nothing,
-    timestamp = :timestamp,
     resolution::Union{Dates.Period, Nothing} = nothing,
 )
     for (k, v) in data
         if v isa DataFrames.DataFrame
-            data[k] = TimeSeries.TimeArray(v; timestamp = timestamp)
+            data[k] = Array(v)
```
daniel-thom, line 25: What happens if the dataframe has a column containing timestamps?

src/deterministic.jl
```diff
@@ -18,26 +18,64 @@ function Deterministic(
...
-            continue
+            data[k] = TimeSeries.values(v)
+            if resolution === nothing
+                resolution = diff(TimeSeries.timestamp(v))[1]
+            end
         else
             try
-                data[k] =
-                    TimeSeries.TimeArray(range(k, length = length(v), step = resolution))
+                data[k] = Vector{Float64}()
```
daniel-thom, line 33: I don't follow this.  Empty vector?  try/catch?

src/deterministic.jl
jd-lara: it's irrelevant in this case. The DataFrame represents a window of the forecast.

src/deterministic.jl
jd-lara: Mishap. Forgot the put the `v` there

src/deterministic.jl
daniel-thom: Then shouldn't you throw an error if the number of columns is greater than 1?

src/deterministic.jl
```diff
@@ -18,29 +18,100 @@ function Deterministic(
...
+        elseif isa(v, Vector{Float64})
             continue
         else
             try
-                data[k] =
-                    TimeSeries.TimeArray(range(k, length = length(v), step = resolution))
+                data[k] = Vector{Float64}(v...)
             catch e
-                throw(ArgumentError("The values in the data dict can't be converted to TimeArrays. Resulting error: $e"))
+                throw(ArgumentError("The values in the data dict are in an invalid format. Resulting error: $e"))
```
daniel-thom, line 37: You're probably better of rethrowing the exception.  What you have could be an error log message.

src/deterministic.jl
```diff
@@ -18,29 +18,100 @@ function Deterministic(
...
-    timestamp = :timestamp,
     resolution::Union{Dates.Period, Nothing} = nothing,
 )
     for (k, v) in data
-        if v isa DataFrames.DataFrame
-            data[k] = TimeSeries.TimeArray(v; timestamp = timestamp)
-        elseif v isa TimeSeries.TimeArray
+        if isa(v, DataFrames.DataFrame)
+            data[k] = Array(v)
+        elseif isa(v, TimeSeries.TimeArray)
```
daniel-thom, line 26: You should probably have a column-length check here also.

daniel-thom (approved):

PR #138, opened 2020-09-30
Time series updates

daniel-thom:
- Reverts a name change.
- Adds user-appropriate docstring for get_time_series.
- Fixes a recently-introduced parsing bug.
- Removes an inadvertent return that messed up a docstring.


jd-lara (approved):
LGTM, thanks for the fix on the docstring

PR #139, opened 2020-09-30
Fix resolution in time series metadata

daniel-thom:


src/time_series_parser.jl
```diff
@@ -23,7 +23,7 @@ mutable struct TimeSeriesFileMetadata
     normalization_factor::Union{String, Float64}
     "Path to the time series data file"
     data_file::String
-    "Resolution of the data being parsed in milliseconds"
+    "Resolution of the data being parsed in seconds"
```
jd-lara, line 26: Need to add a note to myself to document this.

jd-lara (approved):
LGTM

PR #140, opened 2020-09-30
Pre 10 PR 

jd-lara:
Opening for reviews and final fixes. Will be merged when ready. 

- [x] In-memory versions of Probabilistic and Scenarios


src/deterministic.jl
```diff
@@ -0,0 +1,254 @@
...
+Return the forecast window corresponsing to interval index.
+"""
+function get_window(forecast::Deterministic, index::Int)
+    return get_window(forecast, index_to_initial_time(forecast, index))
+end
+
+"""
+Iterate over all forecast windows.
+"""
+function iterate_windows(forecast::Deterministic)
```
daniel-thom, line 227: Needs to be exported

src/deterministic.jl
```diff
@@ -0,0 +1,254 @@
...
+        get_uuid(ts),
+        get_horizon(ts),
+        get_scaling_factor_multiplier(ts),
+    )
+end
+
+"""
+Return the forecast window corresponsing to initial_time.
+"""
+function get_window(forecast::Deterministic, initial_time::Dates.DateTime)
```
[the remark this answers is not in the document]
jd-lara: This should be exported in PSY.

src/deterministic.jl
jd-lara: this too.

src/component.jl
```diff
@@ -1,522 +1,427 @@
...
-    scaling_factors = get_data(forecast)
-    label = get_label(forecast)
-    accessor_func = getfield(mod, Symbol(label))
-    data = scaling_factors .* accessor_func(component)
-    return data
-end
+    name::AbstractString;
+    start_time::Union{Nothing, Dates.DateTime} = nothing,
+    len::Union{Nothing, Int} = nothing,
+    count::Int = 1,
```
daniel-thom, line 114: I'm not sure that this function has ideal default behavior.  If T is a Forecast then it returns only one forecast window.  This could be non-intuitive for users exploring data.  Since the function returns the Forecast object I think it would be better to include all data by default.  If you're concerned about overflowing users' memory, we could put some limit.

The same point applies to all `get_time_series_*` functions.

src/forecasts.jl
```diff
@@ -1,316 +1,91 @@
...
-Return a forecast with only the ending num values.
+Iterate over all forecast windows.
 """
-function tail(forecast::Forecast)
-    return _split_forecast(forecast, TimeSeries.tail(get_data(forecast)))
-end
-
-function tail(forecast::Forecast, num)
-    return _split_forecast(forecast, TimeSeries.tail(get_data(forecast), num))
+function iterate_windows(forecast::Forecast)
```
daniel-thom, line 74: Do we need a method of this function implemented on `Component`?  This method will only work on data that has been loaded into memory.  For example, suppose the user calls `get_time_series` with `count = 10` but there are 100,000 windows.  This function will only iterate over the 10 windows.  Users might want a function that iterates over all windows while only keeping one window in memory at a time.  It would slow because there would be many small disk reads, but it would use very little memory.

src/scenarios.jl
```diff
@@ -0,0 +1,74 @@
+function Scenarios(
+    name::String,
+    data::SortedDict{Dates.DateTime, TimeSeries.TimeArray},
+    scaling_factor_multiplier = nothing,
+)
+    # TODO 1.0: consider determining fields from TimeArrays
+    error("broken")
```
daniel-thom, line 7: Needs implementation

src/scenarios.jl
```diff
@@ -0,0 +1,74 @@
+function Scenarios(
+    name::String,
+    data::SortedDict{Dates.DateTime, TimeSeries.TimeArray},
```
daniel-thom, line 3: All Scenarios constructors need to handle normalization_factor.

src/component.jl
daniel-thom: Discussion from today:  Default behavior for `get_time_series` (only) is to return all data.  In the Hdf storage file we will add a print if the size being read from disk into memory is greater than something like 100 MiB.

daniel-thom:
New interface functions:
`get_time_array(forecast::Forecast, start_time::DateTime)::TimeSeries.TimeArray`
`iterate_time_arrays(forecast::Forecast, start_time::DateTime)`

src/probabilistic.jl
```diff
@@ -0,0 +1,157 @@
+"""
+Construct Probabilistic from a SortedDict of Arrays.
+
+# Arguments
+- `name::AbstractString`: user-defined name
+- `input_data::AbstractDict{Dates.DateTime, Matrix{Float64}}`: time series data.
```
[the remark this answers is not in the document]
kdheepak: Do you mean `Vector{Float64}` here?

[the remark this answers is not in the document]
jd-lara: Probabilistic returns a Matrix of size `(horizon, number_of_percentiles)`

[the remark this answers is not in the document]
daniel-thom: > 
> 
> Probabilistic returns a Matrix of size `(horizon, number_of_percentiles)`

I didn't realize this until just now:  `Matrix` is a alias for `Array{T,2}`.  We should be consistent in use of terms.  The doctrings should match the struct definition.

jd-lara:
@daniel-thom can we consider this ready for review?

src/hdf5_time_series_storage.jl
```diff
@@ -85,140 +85,414 @@ end
...
+    end
+end
+
+function eltype_data(ts::SingleTimeSeries)
+    return eltype(TimeSeries.values(ts.data))
+end
+
+function eltype_data(ts::TimeSeriesData)
+    return eltype(first(values(ts.data)))
+end
```
daniel-thom, line 140: This is broken for DeterministicSingleTimeSeries.  We should break it out by subtype.

src/serialization.jl
```diff
@@ -57,8 +63,59 @@ function serialize(vals::Vector{T}) where {T <: InfrastructureSystemsType}
...
+    if !get(metadata, CONSTRUCT_WITH_PARAMETERS_KEY, false)
+        return base_type
+    end
+
+    # This has several limitations and is only a workaround for PSY.Reserve subtypes.
+    # - each parameter must be in _module
+    # - does not support nested parametrics.
+    # Reserves should be fixed and then we can remove this hack.
+    parameters = [getfield(_module, Symbol(x)) for x in metadata[PARAMETERS_KEY]]
+    return base_type{parameters...}
```
daniel-thom, line 118: @jd-lara You have said that you want to remove parameterization from VariableReserve.  Doing so would require changes in PSI, PSY, and here.  Do you want to do that for v1.0?

jd-lara: we should probably leave it as it is. It is more complex than we need but there isn't anything really broken. We can stage this for the following breaking change.

src/hdf5_time_series_storage.jl
daniel-thom: > 
> 
> @daniel-thom can we consider this ready for review?

I'll make a PR in the early afternoon with some minor fixes identified here.

I think we should do a review of the code coverage report and ensure that no critical logic is uncovered. Do you have time for that?

jd-lara: resolved in #151

claytonpbarrows (approved):
I've generally kept up with the progress, so I'm approving this without thorough review. I'll be updating examples and using over the next week and will submit issues as/if they arise.

jd-lara:
As soon as this PR is ready and merged, I will proceed with a new one to organize the documentation.

src/probabilistic.jl
```diff
@@ -0,0 +1,160 @@
...
+function get_horizon(forecast::Probabilistic)
+    return size(first(values(get_data(forecast))))[1]
+end
+
+eltype_data(forecast::Probabilistic) = eltype_data_common(forecast)
+get_count(forecast::Probabilistic) = get_count_common(forecast)
+get_initial_times(forecast::Probabilistic) = get_initial_times_common(forecast)
+get_initial_timestamp(forecast::Probabilistic) = get_initial_timestamp_common(forecast)
+get_interval(forecast::Probabilistic) = get_interval_common(forecast)
+get_window(forecast::Probabilistic) = get_window_common(forecast)
```
jd-lara, line 157: Duplicate call to get_window

daniel-thom: Fixed

src/probabilistic.jl
```diff
@@ -126,7 +126,7 @@ function ProbabilisticMetadata(time_series::Probabilistic)
     )
 end
 
-convert_data(data::AbstractDict{Dates.DateTime, T}) where {T <: Matrix{<:Real}} =
```
kdheepak: Why isn't this written as:

```suggestion
convert_data(data::AbstractDict{Dates.DateTime, Matrix{T}}) where T <: Real =
```

jd-lara: In Julia when a container is defined inside of a container it needs to be defined as abstract.

daniel-thom: I think the answer to Dheepak's question "tunnel vision" (I wrote this).  His suggestion is good and we should make the change.

src/forecasts.jl
daniel-thom: This has been solved by `ForecastCache`.

daniel-thom (approved):

PR #141, opened 2020-10-01
fix get_time_series_initial_times

jd-lara:
last minute small fix


src/abstract_time_series.jl
```diff
@@ -17,7 +17,7 @@ function get_time_series_initial_times(ts_metadata::ForecastMetadata)
 end
 
 function get_time_series_initial_times(ts_metadata::StaticTimeSeriesMetadata)
-    return Vector{Dates.DateTime}(get_initial_time(ts_metadata))
+    return get_initial_time(ts_metadata)
```
daniel-thom, line 20: I don't follow the reason for these changes.  This function name implies that it will return multiple times.

PR #142, opened 2020-10-02
Fix validation of time series parameters

daniel-thom:


src/system_data.jl
```diff
@@ -549,22 +507,16 @@ end
...
-get_time_series_initial_time(data::SystemData) =
-    get_time_series_initial_time(data.components)
-get_time_series_last_initial_time(data::SystemData) =
-    get_time_series_last_initial_time(data.components)
-get_time_series_horizon(data::SystemData) = get_time_series_horizon(data.time_series_params)
+generate_forecast_initial_times(data::SystemData) =
+    generate_forecast_initial_times(data.time_series_params)
+get_forecast_total_period(data::SystemData) =
+    get_forecast_total_period(data.time_series_params)
+get_forecast_count(data::SystemData) = get_forecast_count(data.time_series_params)
```
daniel-thom, line 514: This term is a bit misleading to me.  Would `get_forecast_window_count` be better?

jd-lara: ok, we could change it

src/system_data.jl
```diff
@@ -549,22 +507,16 @@ end
...
+get_forecast_horizon(data::SystemData) = get_forecast_horizon(data.time_series_params)
+get_forecast_initial_timestamp(data::SystemData) =
```
jd-lara, line 516: This will imply we can close #141 ?

daniel-thom: yes

jd-lara (approved):
LGTM

PR #143, opened 2020-10-03
Remove internal fields from Base.show

daniel-thom:
The goal of this PR is to fix https://github.com/NREL-SIIP/PowerSystems.jl/issues/616.  The code changes include an override of `Base.show` for all subtypes of `InfrastructureSystemsComponent`.  I excluded some internal fields that are of little value to users.  I removed an override of `Base.show` for a vector of components because that shouldn't be needed with this change.  And that overrride was not truncating output.

There can still be a lot of text dumped to the screen when the user calls `get_components(SomeType, system)` (which shows an `Iterators.Flatten`) or calls collect on that result.  Julia should automatically truncate the output in the case of `Iterators.Flatten` and vectors.  The components have a lot of fields, and all of them get printed.  We can discuss eliminating more fields.

@raphaelsaavedra commented in the issue above that he was seeing huge amounts of output when he was printing the output of `get_components`.  When I tested the current code I was seeing truncation occur.  Julia would print a bunch of components then `…` and then a bunch more.  It seemed like a lot of text but it wasn't enough to kill a terminal.  It was printing maybe around 100 components definitely not the thousands that were in my system.  This makes me think that something else was going wrong.  If the issue continues to occur with this new version then we can dig deeper.

### DIFF
--- a/src/probabilistic.jl
+++ b/src/probabilistic.jl
@@ -31,14 +31,15 @@ function Probabilistic(
     end
     data = handle_normalization_factor(input_data, normalization_factor)
 
-
-    return Probabilistic(name,
-                        initial_timestamp,
-                        horizon,
-                        resolution,
-                        percentiles,
-                        data,
-                        scaling_factor_multiplier)
+    return Probabilistic(
+        name,
+        initial_timestamp,
+        horizon,
+        resolution,
+        percentiles,
+        data,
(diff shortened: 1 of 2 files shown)


jd-lara (approved):
LGTM, sorry for pushing to the branch. Was trying to fix an issue with the conflicts

PR #144, opened 2020-10-05
psi fixes

jd-lara:
This change makes the return an Array and not an iterator (or generator) not sure what the return is but not an Array.

### DIFF
--- a/src/component.jl
+++ b/src/component.jl
@@ -37,13 +37,18 @@ end
 
 function _get_columns(start_time, count, ts_metadata::ForecastMetadata)
     offset = start_time - get_initial_timestamp(ts_metadata)
-    interval = get_interval(ts_metadata)
-    index = Int(offset / interval) + 1
+    interval = time_period_conversion(get_interval(ts_metadata))
+    window_count = get_count(ts_metadata)
+    if window_count > 1
+        index = Int(offset / interval) + 1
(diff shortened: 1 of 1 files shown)


src/component.jl
```diff
@@ -208,14 +214,14 @@ function get_time_series_values(
     len::Union{Nothing, Int} = nothing,
     count::Int = 1,
 ) where {T <: TimeSeriesData}
-    return TimeSeries.values(get_time_series_array(
+    return Array{Float64}(TimeSeries.values(get_time_series_array(
```
daniel-thom, line 217: I think there are some more fundamental problems here.  The `get_time_series_*` functions are designed to work with `SingleTimeSeries`.  I'm surprised anything works with `Deterministic`.  We shouldn't be calling `TimeSeries` methods on the data object for `Deterministic` because it is a SortedDict of DateTime to Array.  Also, it will fail if `scaling_factor_multiplier` is set.

This change should be reverted for `SingleTimeSeries` because the data will already be an array.  We need a new layer of internal methods to handle differences between the types.

daniel-thom: Discussion from today:  The `get_time_series_*` functions will not accept count.  For forecasts they will always return one window.

daniel-thom (approved):

PR #145, opened 2020-10-05
Printing improvements

daniel-thom:
Fixes https://github.com/NREL-SIIP/PowerSystems.jl/issues/616 as well as a few miscellaneous items.


jd-lara (approved):
@raphaelsaavedra this should resolve several printing issues

PR #147, opened 2020-10-06
Fix get_time_series functions for Determinstic

daniel-thom:
Probabilistic will be covered in a different PR.


jd-lara (approved):
LGMT. I will merge to test along with the PSI integration.

PR #150, opened 2020-10-09
Updating Deterministic and SingleTimeSeries

sourabhdalvi:
- This PR allows Deterministic and SingleTimeSeries to hold Array{<:Real} or Array{Tuple{Float64,Float64}} or Array{Array{Tuple{Float64,Float64}}}

### DIFF
--- a/src/deterministic.jl
+++ b/src/deterministic.jl
@@ -118,12 +118,7 @@ function Deterministic(
     end
     @assert !isempty(input_data)
 
-    return Deterministic(
-        name,
-        resolution,
-        input_data,
-        scaling_factor_multiplier,
-    )
+    return Deterministic(name, resolution, input_data, scaling_factor_multiplier)
 end
 
 function Deterministic(
@@ -133,18 +128,12 @@ function Deterministic(
     normalization_factor::NormalizationFactor = 1.0,
     scaling_factor_multiplier::Union{Nothing, Function} = nothing,
 )
-
     if !isa(input_data, SortedDict)
         input_data = SortedDict(input_data...)
     end
     @assert !isempty(input_data)
 
-    return Deterministic(
-        name,
-        resolution,
-        input_data,
-        scaling_factor_multiplier,
-    )
+    return Deterministic(name, resolution, input_data, scaling_factor_multiplier)
 end
 
 """
--- a/src/deterministic.jl
+++ b/src/deterministic.jl
@@ -108,22 +108,7 @@ end
 
 function Deterministic(
     name::AbstractString,
-    input_data::AbstractDict{Dates.DateTime, <:Vector{POLYNOMIAL}};
-    resolution::Dates.Period,
-    normalization_factor::NormalizationFactor = 1.0,
-    scaling_factor_multiplier::Union{Nothing, Function} = nothing,
-)
-    if !isa(input_data, SortedDict)
-        input_data = SortedDict(input_data...)
-    end
-    @assert !isempty(input_data)
-
-    return Deterministic(name, resolution, input_data, scaling_factor_multiplier)
-end
-
-function Deterministic(
-    name::AbstractString,
-    input_data::AbstractDict{Dates.DateTime, <:Vector{PWL}};
+    input_data::AbstractDict{Dates.DateTime, <:Vector};
     resolution::Dates.Period,
     normalization_factor::NormalizationFactor = 1.0,
     scaling_factor_multiplier::Union{Nothing, Function} = nothing,
--- a/src/deterministic.jl
+++ b/src/deterministic.jl
@@ -45,7 +45,8 @@ function Deterministic(
     normalization_factor::NormalizationFactor = 1.0,
     scaling_factor_multiplier::Union{Nothing, Function} = nothing,
 )
-    data = SortedDict{Dates.DateTime, Vector{Float64}}()
+    data_type = eltype(TimeSeries.values(first(values(input_data))))
+    data = SortedDict{Dates.DateTime, Vector{data_type}}()
     resolution =
         TimeSeries.timestamp(first(values(input_data)))[2] -
         TimeSeries.timestamp(first(values(input_data)))[1]
@@ -102,6 +103,47 @@ function Deterministic(
         resolution = resolution,
         normalization_factor = normalization_factor,
         scaling_factor_multiplier = scaling_factor_multiplier,
+    )
+end
+
+function Deterministic(
+    name::AbstractString,
+    input_data::AbstractDict{Dates.DateTime, <:Vector{POLYNOMIAL}};
+    resolution::Dates.Period,
+    normalization_factor::NormalizationFactor = 1.0,
+    scaling_factor_multiplier::Union{Nothing, Function} = nothing,
+)
+    if !isa(input_data, SortedDict)
+        input_data = SortedDict(input_data...)
+    end
+    @assert !isempty(input_data)
+
+    return Deterministic(
+        name,
+        resolution,
+        input_data,
+        scaling_factor_multiplier,
+    )
+end
+
+function Deterministic(
+    name::AbstractString,
+    input_data::AbstractDict{Dates.DateTime, <:Vector{PWL}};
+    resolution::Dates.Period,
+    normalization_factor::NormalizationFactor = 1.0,
+    scaling_factor_multiplier::Union{Nothing, Function} = nothing,
+)
+
+    if !isa(input_data, SortedDict)
+        input_data = SortedDict(input_data...)
+    end
+    @assert !isempty(input_data)
+
+    return Deterministic(
+        name,
+        resolution,
+        input_data,
+        scaling_factor_multiplier,
     )
 end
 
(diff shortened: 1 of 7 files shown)


src/deterministic.jl
```diff
@@ -105,6 +106,36 @@ function Deterministic(
...
+    normalization_factor::NormalizationFactor = 1.0,
+    scaling_factor_multiplier::Union{Nothing, Function} = nothing,
+)
+    if !isa(input_data, SortedDict)
+        input_data = SortedDict(input_data...)
+    end
+    @assert !isempty(input_data)
+
+    return Deterministic(name, resolution, input_data, scaling_factor_multiplier)
+end
```
daniel-thom, line 137: Can these be collapsed into one constructor that takes `<:Vector`?

src/hdf5_time_series_storage.jl
```diff
@@ -115,6 +115,30 @@ function serialize_time_series!(
...
+    elseif eltype(first(values(ts.data))) == POLYNOMIAL
+        return "POLYNOMIAL"
+    elseif eltype(first(values(ts.data))) == PWL
+        return "PWL"
+    else
+        error("$(eltype(first(values(ts.data)))) is not supported in forecast data")
+    end
+end
+
+function get_data_type(ts::T) where {T <: StaticTimeSeries}
```
daniel-thom, line 130: StaticTimeSeries is a subtype of TimeSeriesData, so this function should not be needed.

src/hdf5_time_series_storage.jl
```diff
@@ -115,6 +115,30 @@ function serialize_time_series!(
     return
 end
 
+function get_data_type(ts::TimeSeriesData)
```
daniel-thom, line 118: I suggest calling `eltype` only once.  Also, you shouldn't use `.data`. This will not work in my current PR.  Instead, I think we need a function implemented on the concrete types.
```
eltype_data(ts)
```

src/hdf5_time_series_storage.jl
```diff
@@ -162,16 +188,20 @@ function _read_time_series_attributes(
     return data
 end
 
+global type_dict = Dict("CONSTANT" => CONSTANT, "POLYNOMIAL" => POLYNOMIAL, "PWL" => PWL)
```
daniel-thom, line 191:
```suggestion
const _TYPE_DICT = Dict("CONSTANT" => CONSTANT, "POLYNOMIAL" => POLYNOMIAL, "PWL" => PWL)
```

src/hdf5_time_series_storage.jl
```diff
@@ -300,24 +330,122 @@ function deserialize_time_series(
...
-            end
+        data_type = attributes["data_type"]
+        data = get_hdf_array(path, data_type, attributes, rows, columns)
+        new_ts = T(ts_metadata, data)
+    end
+end
+
+function get_hdf_array(
+    path::HDF5.HDF5Group,
+    type::Type{<:CONSTANT},
```
daniel-thom, line 342: I don't see `type` being used.
```suggestion
    ::Type{<:CONSTANT},
```

src/hdf5_time_series_storage.jl
```diff
@@ -300,24 +330,122 @@ function deserialize_time_series(
...
+            data[it] = @view data_read[1:length(rows), i]
         end
+    end
+    return data
+end
 
-        new_ts = T(ts_metadata, data)
+function get_hdf_array(
+    path::HDF5.HDF5Group,
+    type::Type{POLYNOMIAL},
```
daniel-thom, line 365: ditto

src/hdf5_time_series_storage.jl
```diff
@@ -300,24 +330,122 @@ function deserialize_time_series(
...
+        data = get_hdf_array(path, data_type, attributes, rows, columns)
+        new_ts = T(ts_metadata, data)
+    end
+end
+
+function get_hdf_array(
+    path::HDF5.HDF5Group,
+    type::Type{<:CONSTANT},
+    attributes::Dict{String, Any},
+    rows::UnitRange{Int64},
```
daniel-thom, line 344: I recommend never using `Int64` unless the number of bits in the type is actually important.  `Int` or just leave it out altogether.

src/internal.jl
```diff
@@ -10,6 +10,9 @@ abstract type UnitsData end
 end
 
 const _UNIT_SYSTEM_MAP = Dict(string(x) => x for x in instances(UnitSystem))
+const CONSTANT = Real
+const POLYNOMIAL = Tuple{Float64, Float64}
+const PWL = Vector{Tuple{Float64, Float64}}
```
daniel-thom, line 15: Can you put this in `src/common.jl` instead?

test/test_time_series.jl
```diff
@@ -219,6 +277,132 @@ end
...
+    )
+    @test length(IS.get_data(ts3)) == 341
+    #Throws errors
+    @test_throws ArgumentError IS.get_time_series(
+        IS.SingleTimeSeries,
+        component,
+        "test_c";
+        start_time = initial_time,
+        len = 1200,
+    )
```
daniel-thom, line 326: Many of these checks might be redundant, but I'll leave it to you.

daniel-thom (changes requested):
The data transformations look great.  I think this will work well.  I've requested some minor changes.

src/hdf5_time_series_storage.jl
```diff
@@ -115,6 +115,27 @@ function serialize_time_series!(
...
+function get_data_type(ts::TimeSeriesData)
+    data_type = eltype_data(ts)
+    if data_type <: CONSTANT
+        return "CONSTANT"
+    elseif data_type == POLYNOMIAL
+        return "POLYNOMIAL"
+    elseif data_type == PWL
+        return "PWL"
+    else
+        error("$(eltype(first(values(ts.data)))) is not supported in forecast data")
```
daniel-thom, line 127:
```suggestion
        error("$data_type is not supported in forecast data")
```

src/hdf5_time_series_storage.jl
```diff
@@ -300,24 +327,122 @@ function deserialize_time_series(
...
-        if length(columns) == 1
-            data[start_time] = path["data"][rows, columns.start]
-        else
-            data_read = path["data"][rows, columns]
-            for (i, it) in
-                enumerate(range(start_time; length = length(columns), step = interval))
-                data[it] = @view data_read[1:length(rows), i]
-            end
+        data_type = attributes["data_type"]
+        data = get_hdf_array(path, data_type, attributes, rows, columns)
```
daniel-thom, line 332: For a minor improvement I suggest passing the dataset `path["data"]` through stack instead of the HDFGroup.  The caveat is that I don't think you are using the group anywhere.  It will substantially reduce the number of times the hard-coded path name `data` is used.  I would leave it as an untyped parameter `dataset`.

test/test_time_series.jl
```diff
@@ -219,6 +277,86 @@ end
...
+    resolution = Dates.Hour(1)
+    other_time = initial_time + resolution
+    polynomial_cost = repeat([(999.0, 1.0)], 365)
+    data = TimeSeries.TimeArray(
+        range(initial_time; length = 365, step = resolution),
+        polynomial_cost,
+    )
+    data = IS.SingleTimeSeries(data = data, name = "test_c")
+    IS.add_time_series!(sys, component, data)
+    ts1 = IS.get_time_series(
```
daniel-thom, line 296: Is there a reason why you aren't checking the returned data in this test and the next one?  You do check it in later tests.

src/hdf5_time_series_storage.jl
sourabhdalvi: sorry missed that one somehow

src/hdf5_time_series_storage.jl
```diff
@@ -115,6 +115,27 @@ function serialize_time_series!(
     return
 end
 
+function get_data_type(ts::TimeSeriesData)
```
jd-lara, line 118: we should add a comment here that this function is implemented this way to avoid running eval on arbitrary code from the HDF5

sourabhdalvi: ok, I'll add a comment there

src/hdf5_time_series_storage.jl
```diff
@@ -300,24 +327,121 @@ function deserialize_time_series(
...
+end
+
+function get_hdf_array(
+    dataset,
+    ::Type{<:CONSTANT},
+    attributes::Dict{String, Any},
+    rows::UnitRange{Int},
+    columns::UnitRange{Int},
+)
+    data = SortedDict{Dates.DateTime, Array}()
```
jd-lara, line 344: we can do this too. 

```suggestion
function get_hdf_array(
    dataset,
    ::Type{<:T},
    attributes::Dict{String, Any},
    rows::UnitRange{Int},
    columns::UnitRange{Int},
) where T <: CONSTANT
    data = SortedDict{Dates.DateTime, Array{T}}()
```

jd-lara (commented):
With the improvements @daniel-thom suggested this looks great.

src/hdf5_time_series_storage.jl
sourabhdalvi: if I make this change I get a error message saying T not defined when creating the SortedDict

jd-lara: Then let's leave it like it is, interesting error. I got similar strange things with sorted dicts.

jd-lara (approved):
LGTM

daniel-thom (approved):

PR #151, opened 2020-10-14
Minor fixes

daniel-thom:
This mostly just fixes use of `eltype_data` for each subtype of TimeSeriesData.

### DIFF
--- a/src/probabilistic.jl
+++ b/src/probabilistic.jl
@@ -150,6 +150,7 @@ function get_horizon(forecast::Probabilistic)
     return size(first(values(get_data(forecast))))[1]
 end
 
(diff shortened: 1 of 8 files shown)


src/probabilistic.jl
```diff
@@ -28,7 +28,6 @@ function Probabilistic(
     if quantile_count != length(percentiles)
         throw(ArgumentError("The amount of elements in the data doesn't match the length of the percentiles"))
     end
-    data = handle_normalization_factor(input_data, normalization_factor)
```
jd-lara, line 31: Is this not needed anymore?

daniel-thom: This logic is already invoked on line 26.

jd-lara (approved):

PR #152, opened 2020-10-15
scenarios constructors

jd-lara:


src/forecasts.jl
```diff
@@ -120,6 +120,11 @@ function get_window_common(
     end
 
     data = get_data(forecast)[initial_time]
+    if length(size(data)) == 3
+        @assert size(data)[1] == 1
+        data = data[1, :, :]
```
daniel-thom, line 125: Why is this necessary?  A comment would be helpful.  This makes a copy of the data.  You could probably use `@view` instead.

src/forecasts.jl
```diff
@@ -120,6 +120,11 @@ function get_window_common(
...
+    end
+
     if len != horizon
         data = data[1:len]
```
daniel-thom, line 129: Will this be correct for Scenarios?

src/hdf5_time_series_storage.jl
```diff
@@ -480,7 +480,7 @@ function deserialize_time_series(
         data = SortedDict{Dates.DateTime, Array}()
         start_time = attributes["start_time"]
         if length(columns) == 1
-            data[start_time] = path["data"][columns, rows, 1:total_percentiles]
+            data[start_time] = path["data"][1, rows, 1:total_percentiles]
```
daniel-thom, line 483: Why did you need to do this?  Don't we need to be able to read from an offset other than one? The code in the else block uses `columns`.

src/scenarios.jl
```diff
@@ -1,64 +1,91 @@
...
-        scenario_count = scenario_count,
-        data = data,
-        scaling_factor_multiplier = scaling_factor_multiplier,
-    )
+    if !isa(input_data, SortedDict)
+        input_data = SortedDict(input_data...)
+    end
+    data = handle_normalization_factor(input_data, normalization_factor)
+    scenario_count = size(first(values(data)))[2]
+    data = handle_normalization_factor(input_data, normalization_factor)
```
daniel-thom, line 26: This logic already exists on line 24.

src/scenarios.jl
```diff
@@ -1,64 +1,91 @@
+"""
+Construct Scenarios from a SortedDict of Arrays.
+
+# Arguments
+- `name::AbstractString`: user-defined name
+- `input_data::AbstractDict{Dates.DateTime, Matrix{Float64}}`: time series data.
```
daniel-thom, line 6: The function signature uses Array instead of Matrix.  These need to be consistent.

test/test_time_series.jl
```diff
@@ -211,6 +211,36 @@ end
     forecast_retrieved =
         IS.get_time_series(IS.Probabilistic, component, "test"; start_time = initial_time)
     @test IS.get_initial_timestamp(forecast_retrieved) == initial_time
+    t = IS.get_time_series_array(
+        IS.Probabilistic,
+        component,
+        "test";
+        start_time = initial_time,
+    )
+    @test size(t) == (24, 99)
```
daniel-thom, line 220: No data is validated in this test.  Each item is 1.0.  Maybe that's covered elsewhere.  Consider where the data needs to be unique and then verified after the read.

test/test_time_series.jl
```diff
@@ -211,6 +211,36 @@ end
...
+    IS.add_component!(sys, component)
+    forecast = IS.Scenarios(name, d, resolution)
+    IS.add_time_series!(sys, component, forecast)
+    @test IS.has_time_series(component)
+    @test IS.get_initial_timestamp(forecast) == initial_time
+    forecast_retrieved =
+        IS.get_time_series(IS.Scenarios, component, "test"; start_time = initial_time)
+    @test IS.get_initial_timestamp(forecast_retrieved) == initial_time
+    t = IS.get_time_series_array(IS.Scenarios, component, "test"; start_time = initial_time)
+    @test size(t) == (24, 99)
```
daniel-thom, line 243: - The test should use unique data and then verify it after the read.
- The test should include permutations that vary the start_time, length, and count. Consider whether we need something similar to the testset labeled `"Test get_time_series_array Deterministic"`.

daniel-thom (changes requested):

src/scenarios.jl
```diff
@@ -1,64 +1,91 @@
...
+    scenario_count = size(first(values(data)))[2]
+    data = handle_normalization_factor(input_data, normalization_factor)
+
+    return Scenarios(name, resolution, scenario_count, data, scaling_factor_multiplier)
 end
 
 """
-Constructs Scenarios forecast after constructing a TimeArray from initial_time and
-time_steps.
+Construct Scenarios from a Dict of TimeArrays.
```
daniel-thom, line 32: This constructor is not tested.

src/hdf5_time_series_storage.jl
jd-lara: We need to pass a single value. columns here is a range, it should then be `end(columns)` if `columns` is a single one.

daniel-thom (approved):
I left a minor comment that you may want to address before you merge.

src/scenarios.jl
```diff
@@ -1,64 +1,90 @@
...
+        name,
+        data,
+        resolution;
+        normalization_factor = normalization_factor,
+        scaling_factor_multiplier = scaling_factor_multiplier,
+    )
 end
 
-function Scenarios(ts_metadata::ScenariosMetadata, data::Array)
+function Scenarios(ts_metadata::ScenariosMetadata, data::SortedDict{Dates.DateTime, Array})
```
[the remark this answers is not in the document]
jd-lara: It restricts too much the inputs and fails for some cases.

PR #153, opened 2020-10-15
change read-write HDF5 files

jd-lara:


src/hdf5_time_series_storage.jl
```diff
@@ -472,15 +472,17 @@ function deserialize_time_series(
...
-            data_read = path["data"][rows, columns, 1:total_percentiles]
+            data_read = path["data"][1:total_percentiles, rows, columns]
             for (i, it) in enumerate(range(
                 attributes["start_time"];
                 length = length(columns),
                 step = attributes["interval"],
             ))
-                data[it] = @view data_read[1:length(rows), i, 1:total_percentiles]
+                _data = @view data_read[1:total_percentiles, 1:length(rows), i]
+                data[it] = transpose(_data)
```
daniel-thom, line 485: This object will be returned to the user.  It is a rather complex object.  I'm curious as to whether there will be performance implications in how it ends up being used.  Would it be more efficient to call `permute_dims` on the original `data_read` and then return views into it (this copies the data)?  Or maybe `PermutedDimsArray` (this does not copy the data)?  Or is transpose already maximally efficient?

No matter what, it seems like you should perform the "flip" only once.  Right?

jd-lara: The user gets back the corresponding TimeSeriesData object. In the test I ran, Julia is making the conversion when the Probabilistic or Scenarios are created. I made this test and it seems that transpose is already using all the tricks to be efficient 

```Julia
julia> A = [1 2; 3 4]
2×2 Array{Int64,2}:
 1  2
 3  4

julia> b = view(A, :, 1)
2-element view(::Array{Int64,2}, :, 1) with eltype Int64:
 1
 3

julia> transpose(b)
1×2 LinearAlgebra.Transpose{Int64,SubArray{Int64,1,Array{Int64,2},Tuple{Base.Slice{Base.OneTo{Int64}},Int64},true}}:
 1  3
```

daniel-thom: Right.  It gets converted when it gets inserted into the SortedDict.  Per my second comment, wouldn't it be better to transpose only once?

jd-lara: Not sure how PermuteDims handles 3d arrays. Requires testing.

jd-lara: solved via ec3074d

daniel-thom (approved):

PR #154, opened 2020-10-17
Store Deterministic forecasts with floats

daniel-thom:


src/hdf5_time_series_storage.jl
```diff
@@ -126,8 +126,10 @@ function get_data_type(ts::TimeSeriesData)
         return "POLYNOMIAL"
     elseif data_type == PWL
         return "PWL"
+    elseif data_type <: Integer
```
daniel-thom, line 129: We still have a problem converting SingleTimeSeries with integers to floats.  This is a workaround.

test/test_time_series.jl
```diff
@@ -82,21 +78,15 @@ end
     other_time = initial_time + resolution
     name = "test"
     horizon = 24
-    data_vec = Dict(initial_time => ones(horizon), other_time => ones(horizon))
-    data_tuple =
-        Dict(initial_time => tuple(ones(horizon)...), other_time => tuple(ones(horizon)...))
```
daniel-thom, line 87: I don't think we need to support this.  @sourabhdalvi @jd-lara Please comment if you disagree.

src/hdf5_time_series_storage.jl
jd-lara: Doesn't this happen automatically?

test/Project.toml
```diff
@@ -11,6 +11,7 @@ Logging = "56ddb016-857b-54e1-b83d-db4d58db5568"
 Mustache = "ffc61752-8dc7-55ee-8c37-f3e9cdd09e70"
 PrettyTables = "08abe8d2-0d0c-5749-adfa-8a2ac140af0d"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
+Tables = "bd369af6-aec1-5ad0-b16a-f7cc5008161c"
```
jd-lara, line 14: where do we use this?

test/test_time_series.jl
jd-lara: Not anymore, I added it to test the case where we convert to `Vector{Float64}`, given these changes this needs to be eliminated.

jd-lara (changes requested):
The changes look reasonable to me. Needs to eliminate Tables from the toml file

src/hdf5_time_series_storage.jl
daniel-thom: Nope.  CONSTANT is now only Float64

test/Project.toml
daniel-thom: Accident.  Thanks.

src/deterministic.jl
```diff
@@ -205,9 +143,23 @@ function Deterministic(forecast::Deterministic, data::SortedDict{Dates.DateTime,
...
+    SortedDict{Dates.DateTime, Vector{CONSTANT}}(data...)
+convert_data(data::AbstractDict{Dates.DateTime, T}) where {T <: Vector{<:Tuple}} =
+    SortedDict{Dates.DateTime, Vector{POLYNOMIAL}}(data...)
+convert_data(data::AbstractDict{Dates.DateTime, T}) where {T <: Vector{<:Vector{<:Tuple}}} =
+    SortedDict{Dates.DateTime, Vector{PWL}}(data...)
+
+# Workaround for a bug/limitation in SortedDict. If a user tries to construct
+# SortedDict(i => ones(2) for i in 1:2)
+# it won't discern the types and will return SortedDict{Any,Any,Base.Order.ForwardOrdering}
+# https://github.com/JuliaCollections/DataStructures.jl/issues/239
```
daniel-thom, line 156: @jd-lara FYI, you can read the comments in this issue.

src/generated/Probabilistic.jl
```diff
@@ -6,7 +6,7 @@ This file is auto-generated. Do not edit.
         name::String
         resolution::Dates.Period
         percentiles::Vector{Float64}
-        data::SortedDict{Dates.DateTime, Array}
+        data::Union{SortedDict{Dates.DateTime, Matrix{CONSTANT}}, SortedDict{Dates.DateTime, Matrix{POLYNOMIAL}}, SortedDict{Dates.DateTime, Matrix{PWL}}}
```
daniel-thom, line 9: @jd-lara I made comparable changes for Probabilistic.  I suggest that we leave this PR open until I get a chance to fix Scenarios as well.

src/deterministic.jl
jd-lara: Will check

src/hdf5_time_series_storage.jl
jd-lara: We should add a comment on why this is here then.

src/generated/Probabilistic.jl
jd-lara: That's ok, we should target wrapping up Monday though to focus on the pending items in PSY.

daniel-thom: Also, I suspect that deserializing Probabilistic and Scenarios are broken with InMemoryTimeSeriesStorage.  There are no tests for it.  I'll address this.

daniel-thom: @jd-lara I made comparable changes for Scenarios.  I did not make any changes for InMemoryTimeSeriesStorage.  It seems that you did not implement the required deserialize_time_series methods for that struct when you added Probabilistic and Scenarios.  I'll leave this to you.

jd-lara (approved):
LGTM thanks for the fix, this will make the interfaces more consistent

PR #157, opened 2020-10-20
Add TimeSeriesCache to allow efficient iteration

daniel-thom:


src/forecasts.jl
```diff
@@ -126,7 +126,7 @@ function get_window_common(
     end
 
     data = get_data(forecast)[initial_time]
-    if length(size(data)) == 2
+    if ndims(data) == 2
```
jd-lara, line 129: nice change

jd-lara (approved):
Looks great! This will be an important improvement for the performance in forecast read in PSI.

PR #159, opened 2020-10-20
in memory storage fix

jd-lara:


src/in_memory_time_series_storage.jl
```diff
@@ -116,12 +116,12 @@ function deserialize_time_series(
...
 
 function deserialize_time_series(
-    ::Type{<:AbstractDeterministic},
+    ::Type{T},
     storage::InMemoryTimeSeriesStorage,
     ts_metadata::TimeSeriesMetadata,
     rows::UnitRange,
     columns::UnitRange,
-)
+) where T <: TimeSeriesData
```
daniel-thom, line 124: This will fail if the user passes `AbstractDeterministic`.  You could check for that and re-map it to Deterministic.

src/in_memory_time_series_storage.jl
```diff
@@ -116,12 +116,12 @@ function deserialize_time_series(
...
     # TODO 1.0: Much of this will apply to Probabilistic and Scenarios
```
daniel-thom, line 125: This comment can be deleted.

daniel-thom (approved):

PR #160, opened 2020-10-20
Updating TimeSeries methods

sourabhdalvi:
- Adding `get_time_series_by_key`
- Adding new constructors for TimeSeriesKey ( 1. TimeSeriesData, 2.  kwarg for deserialization)
- Adding new `deserialize_struct` for TimeSeriesKey
- Adding test for `get_time_series_by_key`

### DIFF
--- a/src/component.jl
+++ b/src/component.jl
@@ -655,3 +655,38 @@ function _get_time_series_storage(component::InfrastructureSystemsComponent)
 
     return container.time_series_storage
 end
+
+function get_time_series_by_key(
+    key::TimeSeriesKey,
+    component::InfrastructureSystemsComponent;
+    start_time::Union{Nothing, Dates.DateTime} = nothing,
+    len::Union{Nothing, Int} = nothing,
+    count::Union{Nothing, Int} = nothing,
+)
+    return get_time_series(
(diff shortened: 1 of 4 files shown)


src/component.jl
```diff
@@ -655,3 +655,38 @@ function _get_time_series_storage(component::InfrastructureSystemsComponent)
...
+        time_series_metadata_to_data(key.time_series_type),
+        component,
+        key.name,
+        start_time = start_time,
+        len = len,
+        count = count,
+    )
+end
+
+function get_time_series_by_key(
```
daniel-thom, line 676: Is this redundant with the previous function?

sourabhdalvi: This is for the case where we want to get the forecast for service cost, using TimerSeriesKey stored for variable cost.

daniel-thom: We might be missing each other.  The previous function accepts a key, which has `type` and `name`.  This function is identical except that it also takes a `name`.  It seems odd to me that you use the `name` parameter instead of using the `name` inside of the key.  Are the name values identical?  If so, then the function seems redundant.  If they are not the same then I think the interface is a bit confusing.

PR #162, opened 2020-10-21
Add Julia version info to serialization

daniel-thom:


src/InfrastructureSystems.jl
```diff
@@ -7,6 +7,7 @@ import DataFrames
 import Dates
 import JSON3
 import Random
+import Pkg
```
jd-lara, line 10: We should add a note about why this is here since in theory, this is for testing.

jd-lara (approved):

src/InfrastructureSystems.jl
daniel-thom: Not really.  It's used in the serialization process.  Is it OK to assume that a curious reader can search the codebase for "Pkg"?

PR #167, opened 2020-10-30
removing @assert from _get_columns

sourabhdalvi:


src/component.jl
```diff
@@ -52,7 +52,8 @@ function _get_columns(start_time, count, ts_metadata::ForecastMetadata)
     if window_count > 1
         index = Int(offset / interval) + 1
     else
-        @assert interval == Dates.Millisecond(0)
+        # @show interval
```
daniel-thom, line 55: This isn't actually the right fix.  I'll volunteer to make the fix in a separate PR.  But I need some input.  @jd-lara 

This problem happens when the user makes a request like the following:
```
transform_single_time_series!(sys, horizon, interval)
```
where the length the of the original SingleTimeSeries is equal to horizon (there is only one forecast window).  If you construct this exact Deterministic through normal methods then the resulting instance will have an interval of `Dates.Second(0)`.  Technically, this is a ConflictingInputError.  There is not enough data to have a non-zero interval.

Options:
1. In this case the user needs to pass an interval of Dates.Second(0).  The code will currently fail if that happens; that bug can be easily fixed.  If the user passes another interval then we would throw an exception.
2. If the user passes an interval of Dates.Hour(24) with a horizon of 24 where the original length is also 24, we then override the user's request and set it to Dates.Second(0).  This is easy to do.  However, there is a slight oddity in that calling `get_interval()` on the resulting forecasts will return Dates.Second(0).

We should probably go with 2. This may be required for PowerSimulations. The user may not know ahead of time that this scenario exists.

sourabhdalvi: A similar issue cause by this  exists in `get_initial_times` as well.
https://github.com/NREL-SIIP/InfrastructureSystems.jl/blob/63862e656033b015f8a6409361348ee0f8c36592/src/utils/utils.jl#L370

daniel-thom: Right. That has the same root cause. We need to handle interval consistently.  The code assumes that a Deterministic with one window has an interval of 0.

PR #168, opened 2020-10-31
Fix transform of DeterministicSingleTimeSeries with one window

daniel-thom:


src/time_series_parameters.jl
```diff
@@ -71,7 +71,11 @@ function TimeSeriesParameters(
 )
     last_timestamp = initial_timestamp + resolution * len
     last_initial_time = last_timestamp - resolution * horizon
-    count = length(range(initial_timestamp, step = interval, stop = last_initial_time))
+    if interval == Dates.Second(0)
+        count = 1
+    else
+        count = length(range(initial_timestamp, step = interval, stop = last_initial_time))
```
jd-lara, line 77: do we need to allocate the range to calculate this count? isn't easier to do `(last - initial)/interval`

src/utils/utils.jl
```diff
@@ -367,7 +367,7 @@ function get_initial_times(
     if count == 0
         return []
     elseif count == 1
-        @assert interval == Dates.Second(0)
+        @assert interval == Dates.Second(0) "interval=$interval"
```
jd-lara, line 370: I have never seen this syntax of @assert and then string. Does this work?

jd-lara (changes requested):

src/utils/utils.jl
daniel-thom: yes

src/time_series_parameters.jl
daniel-thom: There was actually an off-by-one error here. Fixed.

src/utils/utils.jl
jd-lara: That's cool. It will make it far easier to understand

daniel-thom: What we should do is make a macro that does this for you automatically.

src/time_series_parameters.jl
jd-lara: How about the comment on the count calculation?

daniel-thom: GitHub seems confused on this comment. I did remove the range call, so I'm not sure what you mean. Can you review the code again? I did just add an explanatory comment, but I'm not sure you meant that.

jd-lara (approved):
LGTM

PR #172, opened 2020-11-06
Add assert_op macro

daniel-thom:
This PR adds an assert macro that automatically prints the values that didn't match. A downside to this approach is that the Julia devs say that they plan to compile out assert statements in optimized builds. I don't know how they plan to do that. Use of this macro would not be covered by that.

### DIFF
--- a/src/utils/assert_op.jl
+++ b/src/utils/assert_op.jl
@@ -1,37 +1,39 @@
 """
-Throw an `AssertionError` if `op(exp1, exp2)` is `false`.
+Throw an `AssertionError` if conditions like `op(exp1, exp2)` are `false`, where `op` is a conditional infix operator.
 
 # Examples
+
 ```
 julia> a = 3; b = 4;
 julia> @assert_op a == b
-ERROR: AssertionError: 3 is not isequal 4
+ERROR: AssertionError: 3 == 4
 
-julia> @assert_op a + 3 == b + 4
-ERROR: AssertionError: 6 is not == 8
-
-julia> IS.@assert_op a isequal b
-ERROR: AssertionError: 3 is not isequal 4
+julia> @assert_op a + 3 > b + 4
+ERROR: AssertionError: 6 > 8
 ```
 """
-macro assert_op(exp1, op, exp2)
-    assert_op(exp1, op, exp2)
-end
-
-macro assert_op(expr...)
+macro assert_op(expr)
     assert_op(expr)
 end
 
 function assert_op(expr::Expr)
-    @assert expr.head == :call
-    @assert length(expr.args) == 3
-
-    assert_op(expr.args[2], expr.args[1], expr.args[3])
+    # Only special case expressions of the form `expr1 == expr2`
+    if length(expr.args) == 3 && expr.head == :call
+        return assert_op(expr.args[1], expr.args[2], expr.args[3])
+    else
+        return :(@assert $(expr))
+    end
 end
 
-function assert_op(expr::Tuple{Expr})
-    @assert length(expr) == 1
-    assert_op(expr[1])
+function assert_op(op, exp1, exp2)
+    return :(
+        if !$op($(esc(exp1)), $(esc(exp2)))
+            val1 = $(esc(exp1))
+            val2 = $(esc(exp2))
+            op_str = $(esc(op))
+            throw(AssertionError("$val1 $op_str $val2"))
+        end
+    )
 end
 
 function assert_op(exp1, op, exp2)
--- a/src/utils/assert_op.jl
+++ b/src/utils/assert_op.jl
@@ -35,14 +35,3 @@ function assert_op(op, exp1, exp2)
         end
     )
 end
-
-function assert_op(exp1, op, exp2)
-    return :(
-        if !$op($(esc(exp1)), $(esc(exp2)))
-            val1 = $(esc(exp1))
-            val2 = $(esc(exp2))
-            op_str = $(esc(op))
-            throw(AssertionError("$val1 is not $op_str $val2"))
-        end
-    )
-end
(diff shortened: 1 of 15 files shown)


src/utils/assert_op.jl
```diff
@@ -0,0 +1,31 @@
+"""
+Throw an `AssertionError` if `op(exp1, exp2)` is `false`.
+
+Use only the forms below. `@assert_op a == b` is not currently supported.
```
daniel-thom, line 4: I was not able to easily figure how to support this. If you prefer that this be addressed before we add it to the codebase, I'm OK with it.

jd-lara (approved):
LGTM but it will be good to have a page in docs about it, if you can add it before merging would be great.

daniel-thom:
> 
> 
> LGTM but it will be good to have a page in docs about it, if you can add it before merging would be great.

Sure. If Dheepak also approves then I will do that. I will also go through IS and PSY and apply this in cases where we are calling assert.

src/utils/assert_op.jl
kdheepak: Use this implementation:

```
function assert_op(expr::Expr)
    @assert expr.head == :call
    @assert length(expr.args) == 3

    return :(
        if $(expr.args[1])($(esc(expr.args[2])), $(esc(expr.args[3])))
            return nothing
        else
            name1 = $(string(expr.args[2]))
            name2 = $(string(expr.args[3]))
            val1 = $(esc(expr.args[2]))
            val2 = $(esc(expr.args[3]))
            throw(AssertionError("$name1 = $val1, $name2 = $val2"))
        end
    )

end

macro assert_op(expr)
    return assert_op(expr)
end
```

See output below:

```
$ julia --startup-file=no --project
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.5.2 (2020-09-23)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
|__/                   |

julia> include("test.jl")
@assert_op (macro with 1 method)

julia> a = 3; b = 4;

julia> @assert_op a == b
ERROR: AssertionError: a = 3, b = 4
Stacktrace:
 [1] top-level scope at /Users/USER/gitrepos/pyhelics/test.jl:13

julia> @assert_op (a + 1) == b

julia> @assert_op a + 1 == b

julia> @assert_op a + 3 == b + 4
ERROR: AssertionError: a + 3 = 6, b + 4 = 8
Stacktrace:
 [1] top-level scope at /Users/USER/gitrepos/pyhelics/test.jl:13
```

kdheepak: This will end up with output like `4 = 4` though. We'll have to write more special casing logic to handle that.

```
julia> @assert_op a == 3

julia> @assert_op a == 4
ERROR: AssertionError: a = 3, 4 = 4
Stacktrace:
 [1] top-level scope at /Users/USER/gitrepos/repo/test.jl:13
```

daniel-thom: What do you think of this output instead?
```
julia> @assert_op a == 4
ERROR: AssertionError: 3 is not == 4
Stacktrace:
 [1] top-level scope at /home/dthom/sandboxes/InfrastructureSystems.jl/src/utils/assert_op.jl:49
```

daniel-thom: @kdheepak Thanks for the enhancement. Note that both of our implementations had a bug, though. It was introducing a return into the code. I fixed that here. 

I'll wait for your comment about the new error message.

src/utils/assert_op.jl
```diff
@@ -0,0 +1,46 @@
...
+    assert_op(expr[1])
+end
+
+function assert_op(exp1, op, exp2)
+    return :(
+        if !$op($(esc(exp1)), $(esc(exp2)))
+            val1 = $(esc(exp1))
+            val2 = $(esc(exp2))
+            op_str = $(esc(op))
+            throw(AssertionError("$val1 is not $op_str $val2"))
```
kdheepak, line 43: How about this?

```julia
throw(AssertionError("$op_str($val1, $val2) != true"))
```

Or

```julia
throw(AssertionError("!$op_str($val1, $val2)"))
```

Or

```julia
throw(AssertionError("$val1 !$op_str $val2"))
```

kdheepak: Having the variable name in there seems useful though.

```julia
throw(AssertionError("Assertion failed between $exp1 and $exp2 using $op. $val1 $op_str $val2 is not true"))
```

kdheepak: Screen Shot 2020-11-12 at 10 23 53 AM

kdheepak: Also, there's lots of corner cases this implementation doesn't handle. 

Screen Shot 2020-11-12 at 10 26 05 AM

But I think that is fine, most people using this will be using just the 3 argument expression.

src/utils/assert_op.jl
kdheepak: Resolved in https://github.com/NREL-SIIP/InfrastructureSystems.jl/pull/172/commits/1635b68985e61fbdd100daabecc9263c7deea53a.

kdheepak (approved):

src/utils/assert_op.jl
kdheepak: Resolved in 7973a24.

src/utils/assert_op.jl
```diff
@@ -35,14 +35,3 @@ function assert_op(op, exp1, exp2)
...
-function assert_op(exp1, op, exp2)
-    return :(
-        if !$op($(esc(exp1)), $(esc(exp2)))
-            val1 = $(esc(exp1))
-            val2 = $(esc(exp2))
-            op_str = $(esc(op))
-            throw(AssertionError("$val1 is not $op_str $val2"))
-        end
-    )
-end
```
kdheepak: My bad!

PR #178, opened 2020-11-26
Cb/enable retransform

claytonpbarrows:
This PR addresses an issue where running `transform_single_time_series` more than once fails. The new functionality changes the behavior of some of the negative tests, so they have been adjusted.

There is a related issue of adding a new time_series after `transform_single_time_series` has already been executed. This partially addresses that issue by allowing the user to execute `clear_time_series_transformation!`, then adding the new time series, and then running `transform_single_time_series`. But I can address that more concisely after feedback on this approach.

I believe that @sourabhdalvi, @Nongchao, @ugirumurera, and @andrewrosemberg may all be interested in this.

### DIFF
--- a/src/system_data.jl
+++ b/src/system_data.jl
@@ -311,29 +311,27 @@ function clear_time_series!(data::SystemData)
     reset_info!(data.time_series_params)
 end
 
+"""
+Removes all time series of a particular type from a System.
+
+# Arguments
+- `data::SystemData`: system
+- `type::Type{<:TimeSeriesData}`: Type of time series objects to remove.
+"""
 function remove_time_series!(data::SystemData, ::Type{T}) where T <: TimeSeriesData
     for component in iterate_components_with_time_series(data.components)
         for ts in get_time_series_multiple(component, type = T)
             remove_time_series!(data, typeof(ts), component, get_name(ts))
         end
     end
-end
-#=
-function clear_time_series_transformation!(data::SystemData)
-    for component in iterate_components_with_time_series(data.components)
-        container = get_time_series_container(component)
-        for key in keys(container.data)
-            if key.time_series_type <: ForecastMetadata
-                if remove_time_series_metadata!(component, key.time_series_type, key.name)
-                    error("This should have returned false")
-                end
-            end
-        end
+    counts = get_time_series_counts(data)
+    if counts[2] != 0
+        reset_info!(data.time_series_params)
+    elseif counts[3] !=0
+        reset_info!(data.time_series_params.forecast_params)
     end
-    reset_info!(data.time_series_params.forecast_params)
-    return
 end
-=#
+
 
 """
 Returns an iterator of TimeSeriesData instances attached to the system.
--- a/src/system_data.jl
+++ b/src/system_data.jl
@@ -325,10 +325,12 @@ function remove_time_series!(data::SystemData, ::Type{T}) where {T <: TimeSeries
         end
     end
     counts = get_time_series_counts(data)
-    if counts[2] == 0 # no more time series objects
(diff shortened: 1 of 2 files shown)


src/system_data.jl
```diff
@@ -311,6 +311,21 @@ function clear_time_series!(data::SystemData)
     reset_info!(data.time_series_params)
 end
 
+function clear_time_series_transformation!(data::SystemData)
```
daniel-thom, line 314: If the system has other forecasts (Deterministic, Probabilistic, or Scenarios) then bad things will happen. Calling this function will cause all of their metadata to get deleted. The time series storage (HDF) will still contain the raw data.

You could solve that problem by adding a check for the existence of only SingleTimeSeries and DeterministicSingleTimeSeries.

When we discussed this problem this week I was thinking about a more generic approach:
```
function remove_time_series!(data::SystemData, ::Type{T}) where T <: TimeSeriesData
    for component in iterate_components_with_time_series(data.components)
        for ts in get_time_series_multiple(component, type = T)
            remove_time_series!(data, typeof(ts), component, get_name(ts))
        end
    end
end
```
Example usage: `remove_time_series!(sys, DeterministicSingleTimeSeries)`

I tested this and found an unrelated problem. We need this addition:
```
get_name(value::DeterministicSingleTimeSeries) = get_name(value.single_time_series)
```
The generic solution also needs better handling of resetting of the time_series_params. The current code doesn't handle removing all time series one-by-one. For now we could do what you have here, but with a check. A better fix would be to handle it at a lower level.

I'm fine with either approach.

src/system_data.jl
```diff
@@ -311,6 +311,27 @@ function clear_time_series!(data::SystemData)
...
+"""
+function remove_time_series!(data::SystemData, ::Type{T}) where {T <: TimeSeriesData}
+    for component in iterate_components_with_time_series(data.components)
+        for ts in get_time_series_multiple(component, type = T)
+            remove_time_series!(data, typeof(ts), component, get_name(ts))
+        end
+    end
+    counts = get_time_series_counts(data)
+    if counts[2] == 0 # no more time series objects
+        reset_info!(data.time_series_params)
```
daniel-thom, line 329: This will reset static and forecasts. The main struct defines resolution which is common to both.

src/system_data.jl
```diff
@@ -378,6 +398,8 @@ function transform_single_time_series!(
                 interval,
             )
             check_add_time_series!(data.time_series_params, params)
+            !_is_uninitialized(data.time_series_params.forecast_params) &&
+                remove_time_series!(data, DeterministicSingleTimeSeries)
```
daniel-thom, line 404: You could reach this point if the system has Probabilistic but no DeterministicSingleTimeSeries. Your code will handle it just fine. I suggest adding a test that calls `remove_time_series!` when there are none of that type present. Someone in the future might decide to throw an exception in that case.

daniel-thom (approved):
Looks good

src/system_data.jl
jd-lara: This code seems to have introduced a bug. `!_is_uninitialized(data.time_series_params.forecast_params)` always returns true upon the first call to transform, as a result, remove is always called with the first transformation. Thus removing the forecast parameters just added in the previous call.

src/system_data.jl
```diff
@@ -378,6 +400,8 @@ function transform_single_time_series!(
                 interval,
             )
             check_add_time_series!(data.time_series_params, params)
+            !_is_uninitialized(data.time_series_params.forecast_params) &&
```
jd-lara, line 403: This check is system-wide but conducted inside the for loop for every component.

PR #183, opened 2020-12-08
apply removal correctly

jd-lara:
Fixes a bug introduced in 77d857c1a2c6639e06b082701c6e4f06439eb796 that removes the data for all time forecast always. 

There is one drawback to this approach. If a second call for transform is made and it has inadequate parameters then the correct transformation that was stored before will be eliminated.

### DIFF
--- a/src/system_data.jl
+++ b/src/system_data.jl
@@ -390,6 +390,10 @@ function transform_single_time_series!(
     horizon::Int,
     interval::Dates.Period,
 ) where {T <: DeterministicSingleTimeSeries}
+    @debug data.time_series_params.forecast_params
+    if !is_uninitialized(data.time_series_params.forecast_params)
+        remove_time_series!(data, DeterministicSingleTimeSeries)
+    end
     params = nothing
     for component in iterate_components_with_time_series(data.components)
         if params === nothing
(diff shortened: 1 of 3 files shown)


claytonpbarrows (approved):

src/system_data.jl
```diff
@@ -390,6 +390,10 @@ function transform_single_time_series!(
     horizon::Int,
     interval::Dates.Period,
 ) where {T <: DeterministicSingleTimeSeries}
+    @debug data.time_series_params.forecast_params
```
daniel-thom, line 393: This isn't a great fix because of the reason you mentioned.

The root cause of this issue has two subtle reasons:
1. The call to `check_add_time_series!` mutates the time_series_params. Hence the next call always returns that forecast_params are initialized.
2. Before the changes to remove existing DeterministicTimeSeries the `params` passed to `transform_single_time_series!` were equivalent to the SystemData.time_series_params. Now they are not. We need to pass SystemData.time_series_params.

Replacing this function with the code below should fix the problem in all situations.

```
function transform_single_time_series!(
    data::SystemData,
    ::Type{T},
    horizon::Int,
    interval::Dates.Period,
) where {T <: DeterministicSingleTimeSeries}
    params = nothing
    for component in iterate_components_with_time_series(data.components)
        if params === nothing
            params = get_single_time_series_transformed_parameters(
                component,
                T,
                horizon,
                interval,
            )
            check_add_time_series!(data.time_series_params, params)
            remove_time_series!(data, DeterministicSingleTimeSeries)
        end

        transform_single_time_series!(component, T, data.time_series_params)
    end
end

```

jd-lara: The call `remove_time_series!(data, DeterministicSingleTimeSeries)` is what is causing the problem to begin with. This removes the params created previously.

PR #187, opened 2020-12-24
Updates to Printing and InfrastructureSystemsInternal

sourabhdalvi:


src/utils/print.jl
```diff
@@ -129,6 +129,14 @@ function Base.summary(ist::InfrastructureSystemsComponent)
     return "$(get_name(ist)) ($(typeof(ist)))"
 end
 
+function Base.show(io::IO, ::MIME"text/plain", ist::SystemUnitsSettings)
+    print(io, summary(ist), ":")
+    for (name, field_type) in zip(fieldnames(typeof(ist)), fieldtypes(typeof(ist)))
```
daniel-thom, line 134: You aren't using field_type any longer, so you can remove it.

You probably want to change the name of `ist`.

daniel-thom (approved):

PR #191, opened 2021-01-15
Cb/has component

claytonpbarrows:
add a `has_component` method

### DIFF
--- a/src/components.jl
+++ b/src/components.jl
@@ -146,6 +146,13 @@ function _remove_component!(
 end
 
 """
+Check to see if a component exists.
+"""
+function has_component(::Type{T}, components::Components, name::AbstractString) where T<:InfrastructureSystemsComponent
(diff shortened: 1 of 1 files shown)


src/components.jl
```diff
@@ -145,6 +145,17 @@ function _remove_component!(
...
 
+"""
+Check to see if a component exists.
+"""
+function has_component(
+    ::Type{T},
+    components::Components,
+    name::AbstractString,
+) where {T <: InfrastructureSystemsComponent}
+    !isnothing(get_component(T, components, name))
```
daniel-thom, line 156: I think this should never throw an exception. Here is a proposal:

```suggestion
        !isconcretetype(T) && return !isempty(get_components_by_name(T, components, name))
        !haskey(components.data, T) && return false
        return haskey(components.data[T], name)
```

daniel-thom (approved):

PR #192, opened 2021-01-15
Replace scoped_enum with a better version

### LINKED ISSUE #186: `@scoped_enum` should handle words ending with `y`
I'm opening the issue to track a feature request.

daniel-thom:
Fixes #186 

Please don't merge until we agree on companion PRs in the other repos.

### DIFF
--- a/src/utils/utils.jl
+++ b/src/utils/utils.jl
@@ -192,30 +192,57 @@ function compare_values(x::T, y::U)::Bool where {T, U}
     return x == y
 end
 
+# Copied from https://discourse.julialang.org/t/encapsulating-enum-access-via-dot-syntax/11785/10
 """
-Macro to wrap Enum in a baremodule to keep the top level scope clean.
-The macro name should be singular. The macro will create a module for access that is plural.
+Macro to wrap Enum in a module to keep the top level scope clean.
 
 # Examples
 ```Julia
-@scoped_enum Fruit begin
-    APPLE
-    ORANGE
-end
+julia> @scoped_enum Fruit APPLE = 1 ORANGE = 2
 
-value = Fruits.APPLE
+julia> value = Fruit.APPLE
+Fruit.APPLE = 1
 
-# Usage as a function parameter
-foo(value::Fruits.Fruit) = nothing
+julia> value = Fruit(1)
+Fruit.APPLE = 1
+
+julia> @scoped_enum(Fruit,
+    APPLE = 1,  # comment
+    ORANGE = 2,  # comment
+)
 ```
-
 """
 macro scoped_enum(T, args...)
-    blk = esc(:(baremodule $(Symbol("$(T)s"))
-    using Base: @enum
-    @enum $T $(args...)
-    end))
-    return blk
+    blk = esc(
+        :(
+            module $(Symbol("$(T)Module"))
+            using JSON3
+            export $T
+            struct $T
+                value::Int64
+            end
+            const NAME2VALUE =
+                $(Dict(String(x.args[1]) => Int64(x.args[2]) for x in args))
+            $T(str::String) = $T(NAME2VALUE[str])
+            const VALUE2NAME =
+                $(Dict(Int64(x.args[2]) => String(x.args[1]) for x in args))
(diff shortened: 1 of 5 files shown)


src/utils/utils.jl
```diff
@@ -192,30 +192,56 @@ function compare_values(x::T, y::U)::Bool where {T, U}
...
+            JSON3.StructType(::Type{$T}) = JSON3.StructTypes.StringType()
+
+            function _itr(res)
+                isnothing(res) && return res
+                value, state = res
+                return ($T(value), state)
+            end
+            Base.iterate(::Type{$T}) = _itr(iterate(keys(VALUE2NAME)))
+            Base.iterate(::Type{$T}, state) = _itr(iterate(keys(VALUE2NAME), state))
+            Base.length(::Type{$T}) = mapreduce(x -> 1, +, (x for x in $T))
```
daniel-thom, line 239: @kdheepak Would you like to see any changes here? I did add this line to make it easier to collect all instances.

kdheepak: If you implement `Base.isless(x::T, y::T)` it'll let a user be able to compare and deal with ordering. Do we think that could be useful?

kdheepak: Thoughts on generating this in the macro?

```
Base.convert(::Type{Fruit}, x::Integer) = Fruit(x)
```

It'll allow for functionality like this:

Screen Shot 2021-01-15 at 5 08 33 PM

kdheepak: I'm not sure about using iterate on the `type`. I'd prefer implementing some function (like `pairs`) that the user can then iterate over or call length on.

daniel-thom: This already works
```
julia> Fruit(1)
Fruit.APPLE = 1
```
The struct version does not already work. I'm not opposed to adding it. My initial reaction is that it defeats the point of having strongly-typed enums.

I don't think we need/want ordering. Hopefully, users don't care about comparisons of values. If a user does want ordering, do they want it ordered by name (alphabetical) or by value?

They can always do one of these:
```
sort!([Fruit.APPLE, Fruit.ORANGE], by = x -> x.value)
sort!([Fruit.APPLE, Fruit.ORANGE], by = x -> string(x))
```

daniel-thom: Responding to `iterate`. The original author had that. It allows this:
```
julia> for x in Fruit
       @show x
       end
x = Fruit.ORANGE = 2
x = Fruit.APPLE = 1
```
It did not allow for `[x for x in Fruit]`. Adding `Base.length` makes that work.

`pairs` implies that the user wants the value, right? One of my underlying assumptions is that the value should not matter. Things could be problematic if that does matter and we remove an entry.

kdheepak: Summary, we decided to implement `isless` and not implement `length` or `iterate`. User can use `instances` to get all instances of an Scoped Enum, similar to how they would in `Base`.

daniel-thom: Implemented `isless` and `instances`, deleted `iterate` and `length`, as per our discussion.

jd-lara (approved):
Awesome improvement

test/test_utils.jl
```diff
@@ -28,14 +28,17 @@ end
...
-    ORANGE
-end
+IS.@scoped_enum Fruit APPLE = 1 ORANGE = 2
 
 @testset "Test scoped_enum" begin
-    @test Fruits.APPLE isa Fruits.Fruit
-    @test Fruits.ORANGE isa Fruits.Fruit
+    @test Fruit.APPLE isa Fruit
+    @test Fruit.ORANGE isa Fruit
+    sort([Fruit.ORANGE, Fruit.APPLE]) == [Fruit.APPLE, Fruit.ORANGE]
```
raphaelsaavedra, line 36: Is there a `@test` missing?

```suggestion
    @test sort([Fruit.ORANGE, Fruit.APPLE]) == [Fruit.APPLE, Fruit.ORANGE]
```

daniel-thom: Oops. Thanks!

PR #200, opened 2021-03-03
Fix get_time_series_multiple for Deterministic

daniel-thom:
This fixes a bug that caused the function to return instances of
Deterministic when the caller asked for DeterministicSingleTimeSeries.
There was ambiguity in the mapping of DeterministicMetadata to the
actual time series type. This change makes the mapping explicit.

### DIFF
--- a/src/component.jl
+++ b/src/component.jl
@@ -533,19 +533,14 @@ function get_time_series_multiple(
 
     Channel() do channel
         for key in keys(container.data)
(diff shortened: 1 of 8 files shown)


jd-lara (commented):
@daniel-thom can you provide a bit more context on the issue please?

jd-lara (approved):

PR #207, opened 2021-03-23
Make TimeSeriesCache structs immutable

daniel-thom:


jd-lara (approved):
Just out of curiosity, why is the change necessary? I see you made several fields mutable.

PR #208, opened 2021-03-24
Perform division in Milliseconds

daniel-thom:


src/time_series_cache.jl
```diff
@@ -224,7 +224,9 @@ function ForecastCache(
 
     count = get_count(ts_metadata)
     if start_time != initial_timestamp
-        count -= (start_time - initial_timestamp) ÷ get_interval(ts_metadata)
+        count -=
```
jd-lara, line 227: We have a convert function is IS, it makes it easier in the future to make updates. Just a suggestion.

daniel-thom: I guess you're referring to `time_period_conversion`. In this case we need consistent units in the numerator and denominator to perform the division. The value is not being returned as part of an interface. I would say that `Millisecond(x) ÷ Millisecond(y)` is more clear to readers.

jd-lara (approved):

PR #220, opened 2021-05-03
Add support for masked components

daniel-thom:
This PR enables full support of HybridSystems in PowerSystems. PowerSystems code will "mask" all subcomponents in a HybridSystem. Those subcomponents will still be managed by the system but will not be exposed in functions like `get_components`.

### DIFF
--- a/src/components.jl
+++ b/src/components.jl
@@ -35,7 +35,6 @@ function add_component!(
     components::Components,
     component::T;
     skip_validation = false,
-    deserialization_in_progress = false,
     allow_existing_time_series = false,
 ) where {T <: InfrastructureSystemsComponent}
     component_name = get_name(component)
@@ -51,9 +50,7 @@ function add_component!(
 
     !skip_validation && check_component(components, component)
 
-    if !deserialization_in_progress &&
-       !allow_existing_time_series &&
-       has_time_series(component)
+    if !allow_existing_time_series && has_time_series(component)
         throw(ArgumentError("cannot add a component with time_series: $component"))
     end
 
--- a/src/components.jl
+++ b/src/components.jl
@@ -36,6 +36,7 @@ function add_component!(
     component::T;
     skip_validation = false,
     deserialization_in_progress = false,
+    allow_existing_time_series = false,
 ) where {T <: InfrastructureSystemsComponent}
     component_name = get_name(component)
     if !isconcretetype(T)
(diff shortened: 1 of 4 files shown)


src/system_data.jl
```diff
@@ -314,12 +327,33 @@ function remove_components!(::Type{T}, data::SystemData) where {T}
...
 function clear_time_series!(data::SystemData)
     clear_time_series!(data.time_series_storage)
     clear_time_series!(data.components)
     reset_info!(data.time_series_params)
 end
 
+function iterate_components_with_time_series(data::SystemData)
+    return Iterators.Flatten((
+        iterate_components_with_time_series(data.components),
+        iterate_components_with_time_series(data.masked_components),
```
daniel-thom, line 353: Note that we need to return time series of masked components in order to support `transform_single_time_series!` and `get_time_series_multiple`.

src/components.jl
```diff
@@ -50,7 +51,9 @@ function add_component!(
 
     !skip_validation && check_component(components, component)
 
-    if !deserialization_in_progress && has_time_series(component)
+    if !deserialization_in_progress &&
```
jd-lara, line 54: can we make an intermediate var here to make it easier to read.

jd-lara (approved):

src/components.jl
daniel-thom: Can you be more specific? I'm not following.

jd-lara:
```suggestion
    time_series_condition = !allow_existing_time_series && has_time_series(component)
    if !deserialization_in_progress && time_series_condition
```

jd-lara: Its a minor thing just to understand what the condition is about

daniel-thom: I took the point that it wasn't easy to read and simplified the function.

src/components.jl
```diff
@@ -35,7 +35,7 @@ function add_component!(
     components::Components,
     component::T;
     skip_validation = false,
-    deserialization_in_progress = false,
+    allow_existing_time_series = false,
```
jd-lara, line 38: @daniel-thom This broke the API... we needed to add a deprecation

daniel-thom: As far as I know, PowerSystems is the only module that relies on this, so I think we're safe changing it. It does not affect `PowerSystems.add_components`.

PR #221, opened 2021-05-04
Jd/fix kwargs deprecation

jd-lara:
This fixes the incompatibility introduced in #220

### DIFF
--- a/src/deprecated.jl
+++ b/src/deprecated.jl
@@ -4,7 +4,7 @@ function _add_component_kwarg_deprecation(kwargs)
     if haskey(kwargs, :deserialization_in_progress)
         Base.depwarn("Keyword deserialization_in_progress is deprecated, use allow_existing_time_series instead.", :add_component!)
         kw = Dict(k => v for (k, v) in kwargs if k != :deserialization_in_progress)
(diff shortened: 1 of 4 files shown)


src/deprecated.jl
```diff
@@ -0,0 +1,12 @@
+# 1.4.2 Deprecations
+
+function _add_component_kwarg_deprecation(kwargs)
+    if haskey(kwargs, :deserialization_in_progress)
+        Base.depwarn("Keyword deserialization_in_progress is deprecated, use allow_existing_time_series instead.", :add_component!)
+        kw = Dict(k => v for (k, v) in kwargs if k != :deserialization_in_progress)
+        kw[:allow_existing_time_series] = get(kwargs, :deserialization_in_progress, true)
```
nickrobinson251, line 7: I was confused why this defaults to `true` (when the default, if not passed, is `false`), but i realise this fallback should never be his, but we already checked `haskey`

i.e. since we already checked `haskey`, can this just be
```suggestion
        kw[:allow_existing_time_series] = kwargs[:deserialization_in_progress]
```
?

daniel-thom (approved):

Issue #222, opened 2021-05-04
Breaking change to `add_component!` in v1.4.3


nickrobinson251:
in InfrastructureSystems v1.4.3, a keyword to `add_component!` was renamed from `deserialization_in_progress` -> `allow_existing_time_series`
https://github.com/NREL-SIIP/InfrastructureSystems.jl/pull/220/files#r625452331

AFAICT `add_component!` is public API, so this is a breaking change.

In particular this breaks PowerSystems.jl; it causes a `MethodError` at https://github.com/NREL-SIIP/PowerSystems.jl/blob/v1.4.2/src/base.jl#L383-L386

Which in turn breaks our internal code (which simply calls `add_component!(system, bus)`), leading to PowerSystem's hitting the `MethodError`:

```julia
  MethodError: no method matching add_component!(::InfrastructureSystems.Components, ::Bus; deserialization_in_progress=false, skip_validation=true)
  Closest candidates are:
    add_component!(::InfrastructureSystems.Components, ::T; skip_validation, allow_existing_time_series) where T<:InfrastructureSystems.InfrastructureSystemsComponent at /root/.julia/packages/InfrastructureSystems/q6GNU/src/components.jl:34 got unsupported keyword argument "deserialization_in_progress"
    add_component!(!Matched::InfrastructureSystems.SystemData, ::Any; kwargs...) at /root/.julia/packages/InfrastructureSystems/q6GNU/src/system_data.jl:602
  Stacktrace:
    [1] kwerr(::NamedTuple{(:deserialization_in_progress, :skip_validation), Tuple{Bool, Bool}}, ::Function, ::InfrastructureSystems.Components, ::Bus)
      @ Base ./error.jl:157
    [2] add_component!(data::InfrastructureSystems.SystemData, component::Bus; kwargs::Base.Iterators.Pairs{Symbol, Bool, Tuple{Symbol, Symbol}, NamedTuple{(:deserialization_in_progress, :skip_validation), Tuple{Bool, Bool}}})
      @ InfrastructureSystems ~/.julia/packages/InfrastructureSystems/q6GNU/src/system_data.jl:602
    [3] add_component!(sys::System, component::Bus; skip_validation::Bool, kwargs::Base.Iterators.Pairs{Union{}, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
      @ PowerSystems ~/.julia/packages/PowerSystems/RdpU1/src/base.jl:383
    [4] add_component!
[...]

PR #223, opened 2021-05-04
Catch `ResolverError` to reduce possibility misleading failures from integration tests

nickrobinson251:
When looking into #222, i noticed this package had integration tests with downstream packages - which is great. And they even caught the breaking change in #220 (which were then fixed in #221)

But i noticed one tiny way that might be improved: the integration tests currently allow false negatives, which slightly reduces the signal-noise ratio. This PR changes it so that a PR doesn't fail the integration test if it introduces a breaking change but also bumps to a breaking version; it changes the the Julia code that runs the test to catch a ResolverError and exit with success in that case.

(This is not related to #220, this is just something i saw in passing when investigating if we could further reduce the risk of accidental breaking changes, which hopefully this does by meaning we have even more reason to trust the integration tests)

### DIFF
--- a/.github/workflows/cross-package-test.yml
+++ b/.github/workflows/cross-package-test.yml
@@ -26,9 +26,9 @@ jobs:
         uses: actions/checkout@v2
         with:
           repository: NREL-SIIP/${{matrix.package_name}}.jl
-          path: "${{matrix.package_name}}.jl"
+          path: downstream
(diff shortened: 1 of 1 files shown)


jd-lara (approved):

jd-lara:
@nickrobinson251 thanks. I will implement this in PSY too.

PR #224, opened 2021-05-06
Fix bug in code generator

daniel-thom:
Fixes https://github.com/NREL-SIIP/PowerSystems.jl/issues/733

The code generated duplicate methods if a struct had defaults for all values. I tested the fix on PowerSystems.


daniel-thom:
The cross-package testing failed because IS did not generate the duplicate method. Expected failure.

jd-lara (approved):

PR #227, opened 2021-05-10
Optimize storage of time series

daniel-thom:
This makes two changes:
1. Add option to compress time series data.
2. Change how shared time series references are stored. The earlier
   method caused stranded space.

These changes required the addition of a format version so that we can
maintain backwards compatibility with existing datasets.

I also added some documentation. I added a description of the time series HDF5 format. I also added descriptor rules for our auto-generation code since more people are using that feature.

Regarding compression, I'm thinking that we can expose this with kwargs to the System constructor. We could provide a simple `enable_compression` as well as `compression` that take the `CompressionSettings` struct defined here.

### DIFF
--- a/docs/src/dev_guide/auto_generation.md
+++ b/docs/src/dev_guide/auto_generation.md
@@ -56,7 +56,8 @@ generated. This allows you to enter ``val = MyType(nothing)`` in the REPL and
 see the layout of a struct without worrying about valid values.
 - ``valid_range``: Define this as a Dict with ``min`` and ``max`` and
 InfrastructureSystems will validate any value against that range when you add
-the component to the system.
+the component to the system. Use ``null`` if one doesn't apply, such as if there
+is no max limit.
 - ``validation_action``: Define this as ``error`` or ``warn``. If it is
 ``error`` then InfrastructureSystems will raise an exception if the validation
 code detects a problem. Otherwise, it will log a warning.
--- a/docs/src/dev_guide/auto_generation.md
+++ b/docs/src/dev_guide/auto_generation.md
@@ -25,10 +25,45 @@ reasons to consider using this approach:
 InfrastructureSystems.generate_structs("./src/descriptors/power_system_structs.json", "./src/models/generated")
 ```
 
+## Struct Descriptor Rules
+Each struct descriptor must define the following fields:
+
+- ``struct_name``: Name of struct
+- ``docstring``: Struct docstring
+- ``fields``: Array of struct members. See below for requirements.
+- ``supertype``: Declare the struct with this parent type.
+
+Required fields for each struct member:
+
+- ``name``: Name of field
+- ``data_type``: Type of field
+
+Optional fields for each struct member:
+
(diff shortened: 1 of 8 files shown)


docs/src/dev_guide/auto_generation.md
```diff
@@ -25,10 +25,45 @@ reasons to consider using this approach:
...
+- ``internal_default``: Set to true for non-user-facing fields like
+InfrastructureSystemsInternal that have default values.
+- ``needs_conversion``: Set to true if the getter function needs apply unit
+conversion. The type must implement ``get_value(::InfrastructureSystemsComponent,
+::Type) for this combination of component type and member type.``
+- ``null_value``: Value to indicate the value is null, such as 0.0 for floats.
+If all members in the struct define this field then a "demo" constructor will be
+generated. This allows you to enter ``val = MyType(nothing)`` in the REPL and
+see the layout of a struct without worrying about valid values.
+- ``valid_range``: Define this as a Dict with ``min`` and ``max`` and
```
jd-lara, line 57: Can we add a line to recommend use NULL when the limit doesn't apply. For instance when there is not max.

docs/src/dev_guide/time_series.md
```diff
@@ -38,3 +38,63 @@ deserialize the system.
 1. Add an instance of `InfrastructureSystems.TimeSeriesContainer` to the component struct.
 2. Implement the method `InfrastructureSystems.get_time_series_container` for the
    component. It must return the TimeSeriesContainer object.
+
+
+## Data Format
+Time series arrays are stored in an HDF5 file accoring the format described here.
```
jd-lara, line 44:
```suggestion
Time series arrays are stored in an [HDF5](https://support.hdfgroup.org/HDF5/whatishdf5.html) file according the format described here.
```

docs/src/dev_guide/time_series.md
```diff
@@ -38,3 +38,63 @@ deserialize the system.
...
+- ``compression_enabled``: Specifies whether compression is enabled and will be used for new time series.
+- ``compression_type``: Specifies the type of compression being used.
+- ``compression_level``: Specifies the level of compression being used.
+- ``compression_shuffle``: Specifies whether the shuffle filter is being used.
+
+Each time series array is stored in an HDF5 group named with the array's UUID.
+Each group contains a dataset called ``data`` which contains the actual data.
+Each group also contains a group called ``component_references`` which contains
+an HDF5 attribute for each component reference. The component reference uses the
+format ``<component_uuid>__<time_series_name>``.
```
jd-lara, line 58: label?

jd-lara:
> Regarding compression, I'm thinking that we can expose this with kwargs to the System constructor. We could provide a simple `enable_compression` as well as `compression` that take the `CompressionSettings` struct defined here.

Agreed

docs/src/dev_guide/time_series.md
daniel-thom: `label` isn't used in the code; `name` is.

PR #229, opened 2021-05-11
Add filtering of log messages by group

daniel-thom:
This PR adds a simple solution to the problem where debug log messages from a subject like time series or serialization can overwhelm the log file and you can't debug what you want. Julia already had a solution; we just didn't realize it.

Refer to the changes I made to the time series source file as example usages. If you all like this then I will tag all of the debug messages in IS with appropriate labels.


src/utils/logging.jl
```diff
@@ -241,10 +242,19 @@ MultiLogger([ConsoleLogger(stderr), SimpleLogger(stream)])
 ```
 """
 function MultiLogger(loggers::Array{T}) where {T <: Logging.AbstractLogger}
-    return MultiLogger(loggers, nothing)
+    return MultiLogger(loggers, nothing, Dict{Symbol, Base.LogLevel}())
```
daniel-thom, line 245: I chose `Symbol` instead of `String` because the default value for `group` is the filename containing the log message as a Symbol.

src/utils/logging.jl
```diff
@@ -1,6 +1,14 @@
 
 import Logging
 
+# These use PascalCase to avoid clashing with source filenames.
+const LOG_GROUP_RECORDER = :Recorder
+const LOG_GROUP_SERIALIZATION = :Serialization
+const LOG_GROUP_SYSTEM = :System
+const LOG_GROUP_TIME_SERIES = :TimeSeries
```
daniel-thom, line 8: I think it would be good to use a convention like this to help standardize.

test/runtests.jl
```diff
@@ -86,17 +86,26 @@ function get_logging_level(env_name::String, default)
     return log_level
 end
 
+function get_exclude_from_debug_logging()
+    min_levels = Dict{Symbol, Base.LogLevel}()
+    groups = get(ENV, "ExcludeFromDebugLogging", "")
+    return Dict(Symbol(x) => Logging.Error for x in split(groups, ","))
+end
```
daniel-thom, line 93: Let me know if you have an idea that is less hacky. In my mind this is the easiest way to do what I want to do: exclude noisy log groups while I debug one particular thing. We could do the inverse and have something that says, "enable debug for only group X."

src/utils/logging.jl
jd-lara: Is it possible to define other groups later for PSI in PSI or we need to define them here?

src/utils/logging.jl
```diff
@@ -1,6 +1,14 @@
...
+const LOG_GROUPS =
+    (LOG_GROUP_RECORDER, LOG_GROUP_SERIALIZATION, LOG_GROUP_SYSTEM, LOG_GROUP_TIME_SERIES)
+
```
kdheepak, line 11: Where is `LOG_GROUPS` being used? I couldn't find another instance of it using grep.

test/runtests.jl
claytonpbarrows: If I understand this correctly, it only enables error messages for all groups defined in the "ExcludeFromDebugLogging" environment variable. I tend to like the paradigm where we as for a template that is pre-populated with defaults and then allow the user to modify. So we could do something like:
```
min_levels = get_log_levels(logger)
set_min_level!(logger, :TimeSeries, Logging.Debug)
```

claytonpbarrows: My concern is the user knowing what groups are available

claytonpbarrows (commented):
This is great. I think there are some helper functions that will help make this easy/useful, but I think we should develop those after we get some experience working with it.

test/runtests.jl
kdheepak: If it is an environment variable, I think it should be `SIIP_EXCLUDE_FROM_DEBUG_LOG`, i.e. application name followed by underscore followed by variable name, all upper case.

claytonpbarrows: Ok, now I see `get_min_levels` should serve this purpose.

kdheepak:
I didn't know about `_group`! Neat!

claytonpbarrows:
One more nit:
Is "min_level" the preferred terminology? What about "group_log_level" or "log_group_level"?

src/utils/logging.jl
daniel-thom: We should try to put the common ones here, but I think it would be too onerous to make that a rule. Each package could define its own as needed using a similar convention.

test/runtests.jl
daniel-thom: Agree with @kdheepak that it should be UPPER_CASE if it is an environment variable.

@claytonpbarrows You can do this:
```
ENV["SIIP_EXCLUDE_FROM_DEBUG_LOG"] = "TimeSeries,Serialization,System"
```

I started with an environment variable because that's what we already use for enabling debug logging. A config file may be a better solution. Imagine
```
$ cat log_config.toml
SIIP_CONSOLE_LEVEL = Info
SIIP_FILE_LEVEL = Debug
[Group Levels]
System = Debug
Serialization = Info
TimeSeries = Info
```
Before running tests:
```
ENV["SIIP_LOG_CONFIGURATION"] = "log_config.toml"
```
We could do something similar when running PowerSimulations. If we wanted to be able to change dynamically then we could even have a file watcher detect changes.

src/utils/logging.jl
daniel-thom: > Where is `LOG_GROUPS` being used? I couldn't find another instance of it using grep.

I started making a "disable all predefined" option but then abandoned it. This could still be convenient if you don't remember the formal names or spellings.

daniel-thom:
> One more nit:
> Is "min_level" the preferred terminology? What about "group_log_level" or "log_group_level"?

I started with `set_min_level` because `group` is a parameter name. `set_min_group_level` is probably better. The first argument has type `MultiLogger`, so I think `log` is not necessary.

src/utils/logging.jl
```diff
@@ -1,6 +1,187 @@
...
+const LOG_LEVELS = Dict(
+    "Debug" => Logging.Debug,
+    "Info" => Logging.Info,
+    "Warn" => Logging.Warn,
+    "Error" => Logging.Error,
+)
+
+"""Contains information describing a log event."""
+mutable struct LogEvent
```
daniel-thom, line 33: I had to do some refactoring to get struct composition to work. Sorry for the churn, but most of this has no logic change.

daniel-thom:
Here is a summary of changes I made based on today's discussion:
- Changed terminology: `set_group_level`, `set_group_levels`
- Create function to create default logging config file.
- Consolidated unit test logging customization to use the logging config file. We will do the same in PSY and PSI.
- Updated the docs.

claytonpbarrows (approved):
LGTM

Project.toml
```diff
@@ -17,6 +17,7 @@ Mustache = "ffc61752-8dc7-55ee-8c37-f3e9cdd09e70"
 Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
 PrettyTables = "08abe8d2-0d0c-5749-adfa-8a2ac140af0d"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
+TOML = "fa267f1f-6049-4f14-aa54-33bafae1ed76"
```
jd-lara, line 20: Can you check if the TOML entry requires a compat entry?

jd-lara (commented):
Is the PSY test break expected?

Project.toml
daniel-thom: Turns out that TOML is part of the standard library starting in Julia 1.6. I'll open an issue to remove this once we drop support for 1.5. I added a compat entry here but am not sure if it is required. Let me know if this was wrong.

daniel-thom:
> Is the PSY test break expected?

No. I fixed the bug. Glad we had the cross-package tests.

jd-lara (approved):
Merge whenever you think is ready.

PR #231, opened 2021-05-13
Add filtering of log messages by group - part 2

daniel-thom:
I found a few missing items when implementing log filtering support in PSY.

### DIFF
--- a/src/utils/logging_config.toml
+++ b/src/utils/logging_config.toml
@@ -8,10 +8,10 @@ file_mode = "w+"
 set_global = true
 
 ["group_levels"]
-# Set these to Info to suppress the group when debug logging is enabled.
+# Uncomment these to suppress the group when debug logging is enabled.
 # You can also add the base name of any file (without '.jl') to suppress its debug logging.
-Parsing = "Debug"
-Recorder = "Debug"
-System = "Debug"
-Serialization = "Debug"
-TimeSeries = "Debug"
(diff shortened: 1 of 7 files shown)


src/utils/logging_config.toml
```diff
@@ -8,10 +8,10 @@ file_mode = "w+"
 set_global = true
 
 ["group_levels"]
-# Set these to Info to suppress the group when debug logging is enabled.
+# Uncomment these to suppress the group when debug logging is enabled.
```
jd-lara, line 11: Can you entries to the documentation about this in the PR you have some time? I think this is useful but we need to have a good explanation of this for the users (including us) to actually exploit it.

daniel-thom: Did you see these updates in the last PR?
- https://nrel-siip.github.io/InfrastructureSystems.jl/stable/dev_guide/logging/#Use-Cases
- https://nrel-siip.github.io/InfrastructureSystems.jl/stable/dev_guide/tests/

jd-lara: Yes, I meant on how to modify the TOML file to achieve certain behaviors

daniel-thom: I added more documentation about what we are doing with these groups but am not sure I addressed your concerns. Aren't the comments in this file self-explanatory?

jd-lara (approved):

PR #248, opened 2021-07-23
Don't serialize an empty time series file

daniel-thom:
Fixes https://github.com/NREL-SIIP/PowerSystems.jl/issues/791

This does not require any changes from PowerSystems. However, if user1 serializes a system with no time series with this version of InfrastructureSystems and then shares that system with user2 who has not upgraded, deserialization will fail.


jd-lara (approved):
Thanks, this is a great improvement.

PR #249, opened 2021-07-29
Add time-based log event suppression

daniel-thom:
This PR fixes https://github.com/NREL-SIIP/PowerSystems.jl/issues/775.

Use case:
```
@warn "Noisy log message" maxlog = 10 _suppression_period = 5
```
We take over maxlog if _suppression_period is set. At most, 10 instances of this message will get logged every 5 seconds. If _suppression_period is not set, the standard Julia behavior for maxlog applies: at most 10 log messages will occur for the entire Julia process.

### DIFF
--- a/src/utils/logging.jl
+++ b/src/utils/logging.jl
@@ -1,6 +1,3 @@
-
-import Logging
-
 # These use PascalCase to avoid clashing with source filenames.
 const LOG_GROUP_PARSING = :Parsing
 const LOG_GROUP_RECORDER = :Recorder
@@ -27,7 +24,9 @@ const LOG_LEVELS = Dict(
     "Error" => Logging.Error,
 )
 
-"""Contains information describing a log event."""
+"""
+Contains information describing a log event.
+"""
 mutable struct LogEvent
     file::String
     line::Int
@@ -54,8 +53,11 @@ struct LogEventTracker
     # that method, which results in a warning message from Julia.
     LogEventTracker(events::Dict{Logging.LogLevel, Dict{Symbol, LogEvent}}) = new(events)
 end
-"""Returns a summary of log event counts by level."""
-function report_log_summary(tracker::LogEventTracker)::String
+
+"""
+Returns a summary of log event counts by level.
+"""
+function report_log_summary(tracker::LogEventTracker)
     text = "\nLog message summary:\n"
     # Order by criticality.
     for level in sort!(collect(keys(tracker.events)), rev = true)
@@ -74,7 +76,9 @@ function report_log_summary(tracker::LogEventTracker)::String
     return text
 end
 
-"""Returns an iterable of log events for a level."""
+"""
+Returns an iterable of log events for a level.
+"""
 function get_log_events(tracker::LogEventTracker, level::Logging.LogLevel)
     if !_is_level_valid(tracker, level)
         return []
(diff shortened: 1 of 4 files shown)


src/utils/logging.jl
```diff
@@ -183,6 +193,18 @@ end
...
+the suppression period is 5 seconds. After the period ends, messages will no longer be
+suppressed. Note that this is different from the standard `maxlog` value which applies
+for the duration of your process.
+    
+Customize the period for all events by passing `suppression_period` to this function.
+Customize the period for a specific event by specifying `suppression_period = Y` in your
+log message.
+
+**Note:** Use of log message suppression and the LogEventTracker are not thread-safe.
+Please contact the package developers if you need this functionality.
```
daniel-thom, line 206: Please note this.

src/utils/logging.jl
```diff
@@ -196,10 +214,15 @@ ensure that all events get flushed.
...
 - `set_global::Bool=true`: set the created logger as the global logger
+- `suppression_period::Int`: duration in seconds in which a log event will be supressed if
+   the count exceeds `max_count`.
 
 # Example
 ```Julia
 logger = configure_logging(filename="mylog.txt")
+@info "hello world"
+@info "hello world" max_count = 5
+@info "hello world" max_count = 5 suppression_period = 10
```
daniel-thom, line 225: This should be `_max_count` and `_suppression_period` to avoid collisions with local variables.

jd-lara:
> This PR fixes [NREL-SIIP/PowerSystems.jl#775](https://github.com/NREL-SIIP/PowerSystems.jl/issues/775).
> 
> You may have different opinions on the behavior, so please comment as you see fit. I'm not opposed to making changes. I'll update the documentation after we reach consensus.
> 
> My original plan was to support this:
> 
> ```
> @warn "Noisy log message" maxlog = 50 suppression_timeout = 5s
> ```
> 
> After further consideration I decided that the standard `maxlog` is really kind of pointless because it applies for the duration of the Julia process. If you are running in a REPL you really want suppression to reset after some time. So, I decided on this:
> 
> ```
> @warn "Noisy log message" max_count = 50
> ```
> 
> There is a global `suppression_period` value of 5 seconds that applies to all events. You can customize that for one event like this:
> 
> ```
> @warn "Noisy log message" max_count = 50 suppression_period = 5s
> ```
> 
> Update: There is a bit more complexity. If you create a logger with `InfrastructureSystems.configure_logging`, all of the above applies. Many users of PowerSystems won't do that. They will start Julia and implicitly use the default console logger. In that case the code in this PR that intercepts `max_count` and `suppression_period` won't run. The user will see something like this (note `max_count`):
> 
> ```
> ┌ Warning: Rate 175.0 MW for C5 is larger than the max expected in the range of (min = 47.0, max = 52.0).
> │   max_count = 50
> └ @ PowerSystems ~/sandboxes/PowerSystems.jl/src/utils/IO/branchdata_checks.jl:148
> ```
> 
> If we completely took over `maxlog` in our logger and never defined a custom `suppression_period`, the user would see different behavior between the default Julia logger and our logger, but no "weird" parameters like `max_count` would show up.

We could assume safely that users mostly will limit the use to the default and only change this in some cases. In my experience, if a user wants a specific logger behavior then they control it witj kwargs into the functions.

jd-lara (approved):

PR #252, opened 2021-08-19
Allow override of time series directory with environment variable

daniel-thom:
There is a minor mess that causes CI to fail. Now this PR and https://github.com/NREL-SIIP/PowerSystems.jl/pull/793 are dependent on each other.


jd-lara (approved):
LGTM. This is a good improvement for many of the shared memory applications.

PR #255, opened 2021-08-30
Enhance printing of system and components

daniel-thom:
- Use PrettyTables to print system attributes and components in tables.
- Add show_components to display components of a specific type in a table.
- Add check_time_series_consistency to validate consistency across all SingleTimeSeries.

This may be incompatible with the current PSY version. There will be a companion PR forthcoming.


jd-lara (approved):
This is great. LGTM

PR #259, opened 2021-10-04
Update to latest pretty_table interface

daniel-thom:
The previous code was using a deprecated call.


jd-lara:
@daniel-thom can you add a compat entry to the project.toml file  please? 

```
PrettyTables = "1.2"
```

daniel-thom:
> @daniel-thom can you add a compat entry to the project.toml file please?
> 
> ```
> PrettyTables = "1.2"
> ```

That should not be required. v1.2.0 removed the depecrated methods. The newer methods were already present. I can still do it if you want.

jd-lara:
> > @daniel-thom can you add a compat entry to the project.toml file please?
> > ```
> > PrettyTables = "1.2"
> > ```
> 
> That should not be required. v1.2.0 removed the depecrated methods. The newer methods were already present. I can still do it if you want.

This is for future-proofing since this will allow users to use IS with versions earlier than 1.2 after they remove the deperecation. 

we could do 

 ```
 PrettyTables = "~1.2"
 ```

jd-lara (approved):

PR #260, opened 2021-10-04
formatter fixv2

jd-lara:
@daniel-thom this should prevent dealing with formatter versions in your local environment. I also added new flags.

### DIFF
--- a/.github/workflows/formatter/formatter_code.jl
+++ b/.github/workflows/formatter/formatter_code.jl
@@ -1,3 +1,9 @@
+using Pkg
+Pkg.activate(@__DIR__)
+Pkg.instantiate()
+
+using JuliaFormatter
+
 main_paths = ["./src", "./test"]
 for main_path in main_paths
     for folder in readdir(main_path)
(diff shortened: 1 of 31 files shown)


daniel-thom:
The cases where the formatter "fixed" leaky returns makes the code less clear. Shall I go back through this code and add explicit returns?

src/utils/lazy_dict_from_iterator.jl
```diff
@@ -70,12 +70,12 @@ end
 """Reset the iterator for cases where underlying arrays have changed."""
 function reset_iterator(container::LazyDictFromIterator)
     @debug "reset_iterator"
-    container.state = nothing
+    return container.state = nothing
```
jd-lara, line 73: These ones need manual fixing

.github/workflows/formatter/formatter_code.jl
```diff
@@ -11,7 +17,8 @@ for main_path in main_paths
             remove_extra_newlines = true,
             verbose = true,
             always_for_in = true,
-            whitespace_typedefs = true
+            whitespace_typedefs = true,
+            always_use_return = true
```
daniel-thom, line 21: Just to confirm: this needs to be set to false, right?

jd-lara: correct

PR #266, opened 2021-11-02
Only log progress messages with ProgressLogger

daniel-thom:
If the user enabled debug messages for the console or file, progress
messages would also go to those loggers because the debug level is less
than the progress level.


Project.toml
```diff
@@ -25,7 +25,7 @@ UUIDs = "cf7118a7-6976-5b1a-9a39-7adc72f591a4"
 YAML = "ddb6d928-2868-570f-bddf-ab3f9cf99eb6"
 
 [compat]
-CSV = "~0.8, 0.9"
+CSV = "0.9"
```
jd-lara, line 28: we need to up the documenter csv compat version too.

jd-lara (approved):

PR #267, opened 2021-11-20
Fail transform_single_time_series! on conflicting inputs

daniel-thom:
This handles a case found by @claytonpbarrows where we created an invalid DeterministicSingleTimeSeries. It happened when the first SingleTimeSeries had a length one greater than the second.

### DIFF
--- a/src/system_data.jl
+++ b/src/system_data.jl
@@ -243,6 +243,20 @@ function remove_time_series!(
     return
 end
 
+function remove_time_series!(
+    data::SystemData,
+    component::InfrastructureSystemsComponent,
+    ts_metadata::TimeSeriesMetadata,
+)
+    uuid = get_time_series_uuid(ts_metadata)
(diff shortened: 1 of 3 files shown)


src/system_data.jl
```diff
@@ -369,8 +383,10 @@ Removes all time series of a particular type from a System.
 """
 function remove_time_series!(data::SystemData, ::Type{T}) where {T <: TimeSeriesData}
     for component in iterate_components_with_time_series(data)
-        for ts in get_time_series_multiple(component, type = T)
-            remove_time_series!(data, typeof(ts), component, get_name(ts))
+        for ts_metadata in list_time_series_metadata(component)
```
daniel-thom, line 386: While debugging the issue I noticed that we were unnecessarily reading time series data from the file in order to remove them. I addressed that here.

jd-lara (approved):

PR #268, opened 2021-12-01
Create set_name_internal!

### LINKED ISSUE #230: Remove TOML dependency once support for Julia 1.5 is dropped
TOML is part of the standard library starting in Julia 1.6.

daniel-thom:
This allows consuming packages to block set_name! if necessary.

Fixes #230


src/InfrastructureSystems.jl
```diff
@@ -63,7 +63,8 @@ end
 set_time_series_container!(value::InfrastructureSystemsComponent) = nothing
 
 get_name(value::InfrastructureSystemsComponent) = value.name
-set_name!(value::InfrastructureSystemsComponent, name) = value.name = name
+set_name_internal!(value::InfrastructureSystemsComponent, name) = value.name = name
```
jd-lara, line 66:
```suggestion
function set_name_internal!(value::InfrastructureSystemsComponent, name) 
    value.name = name
    return
    end
```

otherwise the name is returned.

daniel-thom: Good point. Thanks.

Project.toml
```diff
@@ -17,7 +17,6 @@ Mustache = "ffc61752-8dc7-55ee-8c37-f3e9cdd09e70"
 Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
 PrettyTables = "08abe8d2-0d0c-5749-adfa-8a2ac140af0d"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
-TOML = "fa267f1f-6049-4f14-aa54-33bafae1ed76"
```
jd-lara, line 20: we need to leave this entry, it is like the case of Dates

jd-lara (approved):

PR #269, opened 2021-12-07
Make time series simulation field optional

josephmckinsey:
This simulation field seems to have no purpose. We could also remove it very easily (see my branch [remove-simulation-field](https://github.com/josephmckinsey/InfrastructureSystems.jl/tree/remove-simulation-field)).

### DIFF
--- a/src/time_series_parser.jl
+++ b/src/time_series_parser.jl
@@ -81,10 +81,11 @@ function read_time_series_file_metadata(file_path::AbstractString)
                     get(item, "scaling_factor_multiplier", nothing)
                 scaling_factor_multiplier_module =
                     get(item, "scaling_factor_multiplier_module", nothing)
+                simulation = get(item, "simulation", "")
                 push!(
                     metadata,
                     TimeSeriesFileMetadata(;
-                        simulation = item["simulation"],
+                        simulation = simulation,
                         category = item["category"],
                         component_name = item["component_name"],
                         name = item["name"],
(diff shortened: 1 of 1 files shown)


daniel-thom:
Would you mind also setting a default value for the parameter in the TimeSeriesFileMetadata kwarg constructor? We have some exported function that accept this type, and so this change would ensure we never have a problem with the parameter.

daniel-thom (approved):
Looks good. Thanks for making the change.

daniel-thom:
@claytonpbarrows Any updated opinion since yesterday?

josephmckinsey:
> Would you mind also setting a default value for the parameter in the TimeSeriesFileMetadata kwarg constructor? We have some exported function that accept this type, and so this change would ensure we never have a problem with the parameter.

Hopefully we don't have a similar problem with TimeSeriesParsedInfo, since simulation is not a kwarg there so making a default parameter would be a breaking change for any callers.

daniel-thom:
> > Would you mind also setting a default value for the parameter in the TimeSeriesFileMetadata kwarg constructor? We have some exported function that accept this type, and so this change would ensure we never have a problem with the parameter.
> 
> Hopefully we don't have a similar problem with TimeSeriesParsedInfo, since simulation is not a kwarg there so making a default parameter would be a breaking change for any callers.

I think we're OK there because it's only used internally - and only in InfrastructureSystems.

claytonpbarrows:
LGTM, thanks everyone

PR #284, opened 2022-03-21
Disable the formatter for generated files

daniel-thom:


jd-lara (commented):
Do you want to reset the flag for the kwarg space?

jd-lara (approved):

PR #301, opened 2023-01-03
Use proper julia argument order in get_components

### LINKED ISSUE #300: change order of arguments filter function
currently: 

```
IS.get_components(T, sys.data, filter_func)
```

correct 

```
IS.get_components(filter_func, T, sys.data)
```

daniel-thom:
Fixes #300

### DIFF
--- a/src/system_data.jl
+++ b/src/system_data.jl
@@ -712,6 +712,15 @@ function get_component(data::SystemData, uuid::Base.UUID)
 end
 
 function get_components(
+    filter_func::Function,
+    ::Type{T},
+    data::SystemData,
+) where {T}
+    return get_components(T, data.components, filter_func)
+end
+
+# Keep this version to support PowerSystems 1.x.
+function get_components(
     ::Type{T},
     data::SystemData,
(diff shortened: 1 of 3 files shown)


src/system_data.jl
```diff
@@ -711,6 +711,15 @@ function get_component(data::SystemData, uuid::Base.UUID)
...
 
+function get_components(
+    filter_func::Function,
+    ::Type{T},
+    data::SystemData,
+) where {T}
+    return get_components(T, data.components, filter_func)
+end
+
+# Keep this version to support PowerSystems 1.x.
```
daniel-thom, line 722: I'm not sure that we should change IS to 2.0 just for this.

jd-lara: I will leave the choice up to you. At the end of the day we will set compat in PSY 2.0 to a minimum version of IS.

jd-lara (approved):

src/system_data.jl
daniel-thom: If we remove this function then we have to bump to 2.0. I say let's keep backwards compatibility.

PR #307, opened 2023-02-13
Use `Base.@kwdef` for `struct LoggingConfiguration`

kdheepak:
I was looking at how `InfrastructureSystems` handles logging, and thought I'd share a small change as a PR.

Julia has a `Base.@kwdef` that simplifies the keyword only constructors for some structs. I've added an example in this PR for `LoggingConfiguration`. Julia 1.9 exports `@kwdef` so you can even drop the `Base.` if you want to be compat with `"julia" >= v"1.9"`.


daniel-thom (approved):
Cool, thanks. I'll start using this.

PR #322, opened 2024-01-08
Add get_supplemental_attributes for components

daniel-thom:


src/InfrastructureSystems.jl
```diff
@@ -75,7 +75,7 @@ Subtypes may contain time series. Which requires
 All subtypes must include an instance of ComponentUUIDs in order to track
 components attached to each attribute.
 """
-abstract type InfrastructureSystemsSupplementalAttribute <: InfrastructureSystemsType end
+abstract type SupplementalAttribute <: InfrastructureSystemsType end
```
daniel-thom, line 78: I renamed this struct because it needs to get imported and re-exported in PowerSystems. I think we want users to type this name instead.

src/system_data.jl
```diff
@@ -780,11 +780,11 @@ function deserialize(
             attributes_by_uuid[uuid] = attr
         end
     end
-    for component in raw["components"]
-        if haskey(component, "attributes_container")
-            component["attributes_container"] = deserialize(
+    for component in Iterators.Flatten((raw["components"], raw["masked_components"]))
+        if haskey(component, "supplemental_attributes_container")
```
daniel-thom, line 784: One potential problem: this requires that every component that wants supplemental attributes has to use this field name. What do you think about such a requirement?

This solution is advantageous because we can deserialize the attributes here without inspecting types. Another solution is to determine the real types here. That _might_ slow things down some. Another solution is to deserialize attributes in the parent package (PowerSystems). We could think more and test performance impacts.

src/InfrastructureSystems.jl
jd-lara: Ok, the old name was under the premise that the interface in PSY was different and attributed was not used directly

jd-lara (approved):

PR #323, opened 2024-01-09
Updates for supplemental attributes

daniel-thom:
This PR adds these items:
1. Fix handling of `assign_new_uuid`. The function now has to go through the system because of the UUID dictionary that we are managing.
2. Add a function to get all components attached to a supplemental attribute.

### DIFF
--- a/src/component.jl
+++ b/src/component.jl
@@ -269,7 +269,7 @@ function get_time_series_by_key(
     )
 end
 
-function assign_new_uuid!(component::InfrastructureSystemsComponent)
+function assign_new_uuid_internal!(component::InfrastructureSystemsComponent)
     old_uuid = get_uuid(component)
     new_uuid = make_uuid()
     if has_time_series(component)
@@ -277,11 +277,12 @@ function assign_new_uuid!(component::InfrastructureSystemsComponent)
         # There may be duplicates because of transform operations.
         changed_uuids = Set{Tuple{Base.UUID, String}}()
         for (key, ts_metadata) in container.data
-            changed_uuid = (old_uuid, key.name)
+            ts_uuid = get_time_series_uuid(ts_metadata)
+            changed_uuid = (ts_uuid, key.name)
             if !in(changed_uuid, changed_uuids)
                 replace_component_uuid!(
                     container.time_series_storage,
-                    get_time_series_uuid(ts_metadata),
+                    ts_uuid,
                     old_uuid,
                     new_uuid,
                     key.name,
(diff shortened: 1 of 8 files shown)


src/component.jl
```diff
@@ -269,19 +269,20 @@ function get_time_series_by_key(
...
     old_uuid = get_uuid(component)
     new_uuid = make_uuid()
     if has_time_series(component)
         container = get_time_series_container(component)
         # There may be duplicates because of transform operations.
         changed_uuids = Set{Tuple{Base.UUID, String}}()
         for (key, ts_metadata) in container.data
-            changed_uuid = (old_uuid, key.name)
+            ts_uuid = get_time_series_uuid(ts_metadata)
+            changed_uuid = (ts_uuid, key.name)
```
daniel-thom, line 281: This is an existing bug that affects some use case in PowerSystems, like `convert_component!`. Users would know if they experienced it. We could back-port the if necessary. Since no one has reported it, that is probably not needed.

src/system_data.jl
```diff
@@ -851,6 +866,11 @@ end
 get_components_by_name(::Type{T}, data::SystemData, args...) where {T} =
     get_components_by_name(T, data.components, args...)
 
+function get_components(data::SystemData, attribute::SupplementalAttribute)
```
jd-lara, line 869: Are we ok with letting the compiler figure out the `eltype` of the resulting vector or should we add a type annotation?

src/system_data.jl
```diff
@@ -935,8 +955,13 @@ end
 _get_system_basename(system_file) = splitext(basename(system_file))[1]
 _get_secondary_basename(system_basename, name) = system_basename * "_" * name
 
-add_supplemental_attribute!(data::SystemData, component, info; kwargs...) =
-    add_supplemental_attribute!(data.attributes, component, info; kwargs...)
+function add_supplemental_attribute!(data::SystemData, component, info; kwargs...)
```
jd-lara, line 958: Seems a left over use of `info` 
```suggestion
function add_supplemental_attribute!(data::SystemData, component, attribute; kwargs...)
```

src/system_data.jl
```diff
@@ -935,8 +955,13 @@ end
...
+    if isnothing(get_component(typeof(component), data, get_name(component)))
+        throw(ArgumentError("$(summary(component)) is not attached to the system"))
+    end
+
+    return add_supplemental_attribute!(data.attributes, component, info; kwargs...)
```
jd-lara, line 963:
```suggestion
    return add_supplemental_attribute!(data.attributes, component, attribute; kwargs...)
```

src/system_data.jl
daniel-thom: The eltype could be concrete or abstract depending on whether all components are the same type. The compiler should make the eltype the concrete type or first common supertype.

src/component.jl
jd-lara: ok

PR #324, opened 2024-01-26
Use ReTest.jl for tests

daniel-thom:
This PR adopts ReTest.jl in order to make it easier to run subsets of tests.


jd-lara (approved):

test/load_tests.jl
```diff
@@ -0,0 +1,13 @@
+using Revise
```
GabrielKS, line 1: Seems like even if the best practice is to install Revise globally, if our tests are going to depend it we ought to have it in `test/Project.toml`

GabrielKS (commented):
See my initial design-level comments here: https://github.com/NREL-Sienna/PowerSystems.jl/pull/1052#discussion_r1490153760

test/load_tests.jl
jd-lara: This was an oversight. We shouldn't add Revise to the test env

GabrielKS: Revise definitely gets used in this PR though

jd-lara: The default run script for testing in CI doesn't use Revise

PR #326, opened 2024-02-01
Add subsystem interfaces

daniel-thom:


src/subsystems.jl
```diff
@@ -79,7 +79,7 @@ Remove a component from a subsystem.
 
 Throws ArgumentError if the subsystem name or component is not stored.
 """
-function remove_subsystem_component!(
+function remove_component_from_subsystem!(
```
daniel-thom, line 82: This should be symmetric with `add_component_to_subsystem!`.

src/subsystems.jl
```diff
@@ -132,6 +132,35 @@ function get_participating_subsystems(
     return [k for (k, v) in data.subsystems if uuid in v]
 end
 
+"""
+Return true if the component is participating in any subsystems.
+"""
+function is_participating_in_subsystem(
```
daniel-thom, line 138: Question: should we use the term `assigned` instead of `participating`?

jd-lara: I prefer `assigned`

src/subsystems.jl
jd-lara: agreed

src/subsystems.jl
```diff
@@ -124,14 +131,43 @@ end
...
     component::InfrastructureSystemsComponent,
 )
     uuid = get_uuid(component)
     return [k for (k, v) in data.subsystems if uuid in v]
 end
 
+"""
+Return true if the component is assigned to any subsystems.
+"""
+function is_assigned_to_subsystem(
```
jd-lara, line 145:
```suggestion
function is_assigned_to_a_subsystem(
```
or 
```suggestion
function is_assigned_subsystems(
```



Otherwise it could be confused with the query to wether something is attach to a specific subsystem

jd-lara (approved):
A couple of minor suggestions

PR #327, opened 2024-02-09
Add environment variable to override time series directory

daniel-thom:


jd-lara (approved):
I tested in the HPC and this is a saver.

PR #328, opened 2024-02-14
Fix reading of 2d time series data in test code

daniel-thom:
This function, which is only used in tests, was not handling time series data with multiple dimensions.

This is a very old bug. It was triggered by the addition of supplemental attributes and outages to PowerSystem.jl. Tests for those features started using a system with 2-dimensional time series data. That code worked when in-memory time series data used, which happens in PowerSystemCaseBuilder when you set build_system(force_build=true).


jd-lara (approved):
wow... this was indeed an old bug

PR #331, opened 2024-02-22
Variable Cost Refactor Part 1: Function Data

GabrielKS:
The first part of a significant refactor of Sienna cost data structures. See https://github.com/NREL-Sienna/PowerSystems.jl/pull/1056 for more details. This PR consists of a new type `FunctionData`, a new set of concrete subtype structs to represent various functions, and the necessary refactoring to use them rather than `IS.PWL` and such in other InfrastructureSystems data structures. I also performed some minor code deduplication.

IS, PSY, and PSB tests pass when using the appropriate branches of each plus PowerSystemsTestData:
Screenshot of passing tests


src/utils/utils.jl
```diff
@@ -464,8 +496,10 @@ function transform_array_for_hdf(data::Vector{POLYNOMIAL})
     return t_lin_cost
 end
 
-function transform_array_for_hdf(data::SortedDict{Dates.DateTime, Vector{PWL}})
-    quad_cost = hcat(values(data)...)
+function transform_array_for_hdf(
+    data::SortedDict{Dates.DateTime, Vector{PiecewiseLinearPointData}},
+)
+    quad_cost = hcat([get_points.(v) for v in values(data)]...)
```
jd-lara, line 502: Why is this called `quad_cost` for a PWL data? I see that the name is prior to this PR but it seems incorrect.

src/utils/utils.jl
```diff
@@ -481,7 +515,10 @@ function transform_array_for_hdf(data::SortedDict{Dates.DateTime, Vector{PWL}})
     return t_quad_cost
 end
 
-function transform_array_for_hdf(data::Vector{PWL})
+# TODO: old code here does not properly handle data with different numbers of points
```
jd-lara, line 518: This is ok. It was meant like that for proper storage in the HDF5 file

jd-lara (approved):
This PR looks ok to me, there is some pending work for the MarketBid Cost that requires scoping but we already decided to refactor that at a later stage.

src/utils/utils.jl
GabrielKS: It does seem incorrect. I'll change it

PR #332, opened 2024-02-26
Variable Cost Refactor Part 2: PowerSimulations

GabrielKS:
The second part of a significant refactor of Sienna cost data structures (part 1 is [here](https://github.com/NREL-Sienna/PowerSystems.jl/pull/1056#issuecomment-1959057500)). The main focus here is to bring PowerSimulations.jl up to date with the elimination of `VariableCost` in favor of `FunctionData`. That entails these changes in InfrastructureSystems:

- Make `PiecewiseLinearPointData` use `NamedTuples` so it's even harder to mix up x and y
- Add the notion of "raw data" to support storing data from `FunctionData` in other containers, like the ones in PowerSimulations

On the latter point — does this functionality belong in PowerSimulations?

Tests pass:
Screenshot 2024-02-26 at 4 32 27 AM

### DIFF
--- a/src/function_data.jl
+++ b/src/function_data.jl
@@ -64,15 +64,40 @@ representation of cost functions where the points store quantities (x, y), such 
 The curve starts at the first point given, not the origin.
 
 # Arguments
- - `points::Vector{Tuple{Float64, Float64}}`: the points (x, y) that define the function
+ - `points::Vector{@NamedTuple{x::Float64, y::Float64}}`: the points that define the function
 """
 struct PiecewiseLinearPointData <: FunctionData
-    points::Vector{Tuple{Float64, Float64}}
+    points::Vector{@NamedTuple{x::Float64, y::Float64}}
 
-    function PiecewiseLinearPointData(points)
+    function PiecewiseLinearPointData(points::Vector{<:NamedTuple{(:x, :y)}})
         _validate_piecewise_x(first.(points))
         return new(points)
     end
+end
+
+function PiecewiseLinearPointData(points::Vector{<:NamedTuple})
+    throw(
+        ArgumentError(
+            "If constructing PiecewiseLinearPointData with NamedTuples, points must have type Vector{<:NamedTuple{(:x, :y)}}; got $(typeof(points))",
+        ),
+    )
+end
+
+function _convert_to_xy(point)
+    # Need to be able to handle dicts for deserialization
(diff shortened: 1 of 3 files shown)


GabrielKS:
The group of four PRs this is a part of:

1. https://github.com/NREL-Sienna/InfrastructureSystems.jl/pull/332
2. https://github.com/NREL-Sienna/PowerSystems.jl/pull/1059
3. https://github.com/NREL-Sienna/PowerSystemCaseBuilder.jl/pull/73
4. https://github.com/NREL-Sienna/PowerSimulations.jl/pull/1061

src/function_data.jl
```diff
@@ -64,24 +64,49 @@ representation of cost functions where the points store quantities (x, y), such
 The curve starts at the first point given, not the origin.
 
 # Arguments
- - `points::Vector{Tuple{Float64, Float64}}`: the points (x, y) that define the function
+ - `points::Vector{@NamedTuple{x::Float64, y::Float64}}`: the points that define the function
 """
 struct PiecewiseLinearPointData <: FunctionData
-    points::Vector{Tuple{Float64, Float64}}
+    points::Vector{@NamedTuple{x::Float64, y::Float64}}
```
jd-lara, line 70: we should create and Alias for this similarly to how it is done in PSY with MinMax adding something like  

```
const TWO_DIM_COORDINATES = NamedTuple.... 
```

src/function_data.jl
```diff
@@ -64,24 +64,49 @@ representation of cost functions where the points store quantities (x, y), such
...
 
+function PiecewiseLinearPointData(points::Vector{<:NamedTuple})
+    throw(
+        ArgumentError(
+            "If constructing PiecewiseLinearPointData with NamedTuples, points must have type Vector{<:NamedTuple{(:x, :y)}}; got $(typeof(points))",
+        ),
+    )
+end
+
+function _convert_to_xy(point)
```
jd-lara, line 86:
```suggestion
function _convert_to_named_tuple(point)
```

jd-lara (changes requested):
Use Alias instead of declaring the NamedTuple everywhere.

src/function_data.jl
GabrielKS: How's `_convert_to_xy_coords` given the name of my new `NamedTuple` alias?

jd-lara (approved):

PR #333, opened 2024-02-27
Add functionality to serialize a system without time series

daniel-thom:
There are three sets of changes here:
1. Add a `to_dict` method so that the IS system and components can be serialized to a JSON string without the time series data.
2. Remove serialization of a system’s validation descriptors. This is a breaking change in that any custom packages that were defining their own validation descriptors will need to pass them at deserialization time (there probably aren’t any). This actually fixes an issue where the descriptors were not upgraded by PSY during an upgrade across major versions.
3. Update the test code that compares objects recursively to exclude comparison of time series fields that will not be valid in this scenario.

### DIFF
--- a/src/components.jl
+++ b/src/components.jl
@@ -356,14 +356,20 @@ function set_name!(
     return
 end
 
-function compare_values(x::Components, y::Components; compare_uuids = false)
+function compare_values(
+    x::Components,
+    y::Components;
+    compare_uuids = false,
+    exclude = Set{Symbol}(),
+)
     match = true
     for name in fieldnames(Components)
+        name in exclude && continue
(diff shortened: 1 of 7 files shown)


jd-lara (commented):
If I follow the logic behind this case, then the objective is to not check for UUIDs in the validation step

daniel-thom:
> If I follow the logic behind this case, then the objective is to not check for UUIDs in the validation step

In the case where we serialize/de-serialize the system to/from a JSON string, losing the time series, we need to be able to exclude comparison of time series fields while traversing the components. The new `exclude` parameter allows us to skip custom fields.

jd-lara (approved):

PR #340, opened 2024-03-09
Resolve nameof vs strip_module_name

daniel-thom:
This is a follow-on to #339. There were cases where we were using `nameof` (which removes parameteric types) where we should have used `strip_module_name` (which preserves parametric types). The best example is when printing a table of component counts. We were showing `VariableReserve` twice instead of showing both `VariableReserve{ReserveDown}` and `VariableReserve{ReserveUp}`.

### DIFF
--- a/src/utils/utils.jl
+++ b/src/utils/utils.jl
@@ -60,7 +60,19 @@ function supertypes(::Type{T}, types = []) where {T}
 end
 
 """
-Strips the module name off of a type.
+Strips the module name off of a type. This can be useful to print types as strings and
+receive consistent results regardless of whether the user used `import` or `using` to
+load a package.
+
+Unlike Base.nameof, this function preserves any parametric types.
+
+# Examples
+```julia-repl
+julia> strip_module_name(PowerSystems.RegulationDevice{ThermalStandard})
+"RegulationDevice{ThermalStandard}"
+julia> string(nameof(PowerSystems.RegulationDevice{ThermalStandard}))
+"RegulationDevice"
+```
 """
 function strip_module_name(name::String)
     index = findfirst(".", name)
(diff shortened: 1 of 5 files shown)


src/utils/utils.jl
```diff
@@ -297,7 +299,7 @@ function forward(sender::Tuple{Type, Symbol}, ::Type, method::Method)
     # Assert that function name always starts with `get_*`
     "`forward` only works for accessor methods that are defined as `get_*` or `set_*`"
     @assert startswith(string(method.name), r"set_|get_")
-    sender_type = "$(parentmodule(sender[1])).$(strip_module_name(sender[1]))"
+    sender_type = string(sender[1])
```
daniel-thom, line 302: Tell me if I’m missing something, but I think this should be sufficient.

jd-lara: We can test it with the printing of the RTS

jd-lara (approved):

GabrielKS:
These were the "more intricacies to figure out" I was talking about :)

GabrielKS:
I had assumed that surely there was a way to get the parameters of a type in general, but it seems like maybe the only ways to do this without string manipulation use non-public interface.

GabrielKS (approved):

daniel-thom:
> I had assumed that surely there was a way to get the parameters of a type in general, but it seems like maybe the only ways to do this without string manipulation use non-public interface.

Like this? https://github.com/NREL-Sienna/InfrastructureSystems.jl/blob/main/src/serialization.jl#L115

GabrielKS:
> > I had assumed that surely there was a way to get the parameters of a type in general, but it seems like maybe the only ways to do this without string manipulation use non-public interface.
> 
> Like this? https://github.com/NREL-Sienna/InfrastructureSystems.jl/blob/main/src/serialization.jl#L115

I was under the impression that `.parameters` was not considered public interface.

PR #341, opened 2024-03-16
Allow bulk writing and reading of time series

daniel-thom:
This is at least a partial fix of https://github.com/NREL-Sienna/PowerSystems.jl/issues/1076. The current behavior is to open and close the file once per addition of a time series array. The open/close has non-trivial overhead.

This PR allows the user to open the store for bulk writes and reads.

There may be other performance issues experienced by the submitter. I’d like to see what happens with this change by itself.  

This PR is targeted at main. We could easily backport it to the current IS/PSY tags.

### DIFF
--- a/src/system_data.jl
+++ b/src/system_data.jl
@@ -108,6 +108,16 @@ function SystemData(
     )
 end
 
+function open_time_series_store!(
+    func::Function,
+    data::SystemData,
+    mode = "r",
+    args...;
+    kwargs...,
+)
+    open_store!(func, data.time_series_storage, mode, args...; kwargs...)
+end
+
 """
 Adds time_series from a metadata file or metadata descriptors.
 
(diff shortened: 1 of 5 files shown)


src/system_data.jl
```diff
@@ -144,10 +154,12 @@ function add_time_series_from_file_metadata!(
     file_metadata::Vector{TimeSeriesFileMetadata};
     resolution = nothing,
 ) where {T <: InfrastructureSystemsComponent}
-    cache = TimeSeriesParsingCache()
-    for metadata in file_metadata
-        if resolution === nothing || metadata.resolution == resolution
-            add_time_series_from_file_metadata_internal!(data, T, cache, metadata)
+    open_time_series_store!(data, "r+") do
```
daniel-thom, line 157: This will speedup time series additions from our JSON/CSV files, which is used by the table data parser.

jd-lara (approved):
Looks good to me. We can now support a slightly different workflow. I think we should keep it in main for IS 2.0.

PR #343, opened 2024-03-22
Share time series data on a component across multiple fields

daniel-thom:
This PR adds new time series data constructors that enable time series instances with different names to share the same data. The targeted use case is one where a generator wants to use the same time series data for the fields active_power and reactive_power.


jd-lara (approved):

jd-lara:
@GabrielKS we might need to make cases in PSB that include this.

PR #344, opened 2024-04-03
Second Cost Refactor: Companion PR (+add serialization for `@scoped_enum`, add `NotImplementedError`)

GabrielKS:
Draft for now, this is the companion to https://github.com/NREL-Sienna/PowerSystems.jl/pull/1071.


GabrielKS:
@daniel-thom thoughts on automatic `@scoped_enum` serialization support along [these](https://github.com/NREL-Sienna/InfrastructureSystems.jl/pull/344/commits/3b11c1c13a25fd88e5937d44f56945d205b83842) lines?

daniel-thom:
> @daniel-thom thoughts on automatic `@scoped_enum` serialization support along [these](https://github.com/NREL-Sienna/InfrastructureSystems.jl/pull/344/commits/3b11c1c13a25fd88e5937d44f56945d205b83842) lines?

NIce. We can remove some similar code from PSY.

jd-lara (commented):
Is this PR still in draft or is it ready for review?

GabrielKS:
> Is this PR still in draft or is it ready for review?

I think it's ready for review, assuming we don't end up needing to make any more changes here related to the market bid cost.

src/function_data.jl
```diff
@@ -34,51 +38,38 @@ get_quadratic_term(fd::QuadraticFunctionData) = fd.quadratic_term
...
-"""
-struct PolynomialFunctionData <: FunctionData
-    coefficients::Dict{Int, Float64}
-end
-
-get_coefficients(fd::PolynomialFunctionData) = fd.coefficients
-
 function _validate_piecewise_x(x_coords)
-    # TODO currently there exist cases where we are constructing a PiecewiseLinearPointData
+    # TODO currently there exist cases where we are constructing a PiecewiseLinearData
```
jd-lara, line 42: What about this TODO

src/function_data.jl
```diff
@@ -254,27 +219,38 @@ deserialize(::Type{FunctionData}, val::Dict) =
 Get a bare numerical representation of the data represented by the FunctionData
 """
 function get_raw_data end
-get_raw_data(fd::LinearFunctionData) = get_proportional_term(fd)
+get_raw_data(fd::LinearFunctionData) = (get_proportional_term(fd), get_constant_term(fd))
```
jd-lara, line 222: get_raw_data here is not type stable. Test it with a type annotation inside a function using @code_warntype

src/hdf5_time_series_storage.jl
```diff
@@ -667,72 +656,48 @@ end
...
+)
+    length_req = fieldcount(get_raw_data_type(T))
+    (size(data)[end] == length_req) || throw(
+        ArgumentError(
+            "Last dimension of data must have length $length_req, got size $(size(data))",
+        ),
+    )
+    dims_to_keep = Tuple(1:(ndims(data) - 1))
+    # Pop off the last dimension and call the constructor on that data
+    return map(x -> T(x...), eachslice(data; dims = dims_to_keep))
```
jd-lara, line 696: Test the performance of this. This is a function that would be called a lot and I am not sure if this is better than the previous approach allocating the array and filling it.

src/utils/utils.jl
```diff
@@ -498,13 +478,13 @@ function transform_array_for_hdf(
     return t_lin_cost
 end
 
-function transform_array_for_hdf(data::Vector{QuadraticFunctionData})
-    all(get_constant_term.(data) .== 0) ||
-        throw(ArgumentError("Not yet implemented for nonzero constant term"))
-    data = [(get_quadratic_term(q), (get_proportional_term(q))) for q in data]
+function transform_array_for_hdf(
```
jd-lara, line 481: You might want to implement this function closer to the previous method. Even when there is code repetition some of the variables' types won't be determined at run time.

src/utils/utils.jl
```diff
@@ -535,7 +515,7 @@ end
 
 # TODO: old code here does not properly handle data with different numbers of points
```
jd-lara, line 516: What about these TODOs

jd-lara (commented):
I have some concerns on performance critical portions of the code.

src/function_data.jl
GabrielKS: Indeed — here's a MWE, for the record:
```julia
@code_warntype IS.transform_array_for_hdf(SortedDict(DateTime("2023-01-01") => [LinearFunctionData(1.0)]))
```
I'll work on fixing this.

src/common.jl
```diff
@@ -22,6 +22,21 @@ struct HashMismatchError <: Exception
     msg::AbstractString
 end
 
+"""
+Indicate that the feature at hand happens to not be implemented for the given data even
+though it could be. If it is a category mistake to imagine this feature defined on that
+data, use another exception, like `TypeError` or `ArgumentError`.
+"""
+struct UnimplementedError <: Exception
```
daniel-thom, line 30: At risk of nit-picking a good idea unnecessarily, would you mind copying Python's `NotImplementedError`? I write that constantly in my Python work and would like the consistency.

GabrielKS: Done

src/function_data.jl
jd-lara: Open an issue to tackle this so we can merge this + the PSY companion to continue on to the PSI work.

jd-lara:
@GabrielKS please merge as soon as a tracking issue with the remaining TODO's is open please. Same for https://github.com/NREL-Sienna/PowerSystems.jl/pull/1071.

src/utils/utils.jl
GabrielKS: Addressed (won't fix the handling data with different numbers of points but added a validation check for it)

GabrielKS:
Tracking issue is https://github.com/NREL-Sienna/InfrastructureSystems.jl/issues/352, I'll merge this as soon as I've verified tests pass.

src/deterministic.jl
```diff
@@ -216,21 +216,23 @@ function Deterministic(
...
+# If values are no more specific than Any, assume CONSTANT
+convert_data(data::AbstractDict{<:Any, Any}) =
+    SortedDict{Dates.DateTime, Vector{CONSTANT}}(data...)
+
+# If values are more specific, don't assume CONSTANT but do upgrade some types
+convert_data(data::AbstractDict{<:Any, Vector{T}}) where {T} =
+    SortedDict{Dates.DateTime, Vector{T}}(data...)
+
+# If everything is fully specified, pass through
+convert_data(data::SortedDict{Dates.DateTime, Vector}) = data
```
daniel-thom, line 235: Are you sure that this works? I'm uncertain about just `Vector`. I see that it isn't tested.

src/hdf5_time_series_storage.jl
```diff
@@ -667,114 +672,58 @@ end
...
-            t_data[r, c] = tuple_array
-        end
-    end
-    return PiecewiseLinearPointData.(t_data)
+function retransform_hdf_array(
+    data::Array,
+    T::Union{Type{LinearFunctionData}, Type{QuadraticFunctionData}},
+)
+    length_req = fieldcount(get_raw_data_type(T))
+    (size(data)[end] == length_req) || throw(
```
daniel-thom, line 701: I don't find anything objectionable about this pattern. However, I do want consistency in our error checks. So, either we always do it this way or always do it the other.
```
size(data)[end] != length_req && throw
```
or 
```
if size(data)[end] != length_req
    throw ...
end
```

src/utils/utils.jl
```diff
@@ -455,106 +461,82 @@ function transform_array_for_hdf(data::Vector{<:Real})
...
+    for r in 1:rows, c in 1:cols, t in 1:n_points
+        t_quad_cost[r, c, t, :] = collect(quad_cost[r, c][t])
     end
     return t_quad_cost
 end
 
-transform_array_for_hdf(
-    data::SortedDict{Dates.DateTime, Vector{T}}) where {T <: FunctionData} =
-    throw(ArgumentError("Not currently implemented for $T"))
+transform_array_for_hdf(::Vector{T}) where {T <: FunctionData} =
```
daniel-thom, line 538: Same comment as elsewhere...we can rely on Julia.

src/function_data.jl
GabrielKS: Done, https://github.com/NREL-Sienna/InfrastructureSystems.jl/issues/352

src/function_data.jl
```diff
@@ -34,51 +38,35 @@ get_quadratic_term(fd::QuadraticFunctionData) = fd.quadratic_term
...
 
-    function PiecewiseLinearPointData(points::Vector{<:NamedTuple{(:x, :y)}})
+    function PiecewiseLinearData(points::Vector{<:NamedTuple{(:x, :y)}})
         _validate_piecewise_x(first.(points))
         return new(points)
     end
 end
 
-function PiecewiseLinearPointData(points::Vector{<:NamedTuple})
+function PiecewiseLinearData(points::Vector{<:NamedTuple})
```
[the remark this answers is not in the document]
GabrielKS: Without this method, the default error would be:
```julia
julia> IS.PiecewiseLinearData([(a=1, b=1), (a=2, b=3)])
ERROR: type NamedTuple has no field x
```
which doesn't seem super informative

src/hdf5_time_series_storage.jl
GabrielKS: Is it the short-circuiting operator rather than the `if` block you take issue with or the fact that I'm using `x || throw` rather than `!x && throw`? Happy to change these to the existing convention.

daniel-thom: I'm fine with leaving this as-is.

GabrielKS: Removed all the occurrences of the `||` pattern

src/utils/utils.jl
GabrielKS: Removed

PR #346, opened 2024-04-09
Support repeated reads of cached timestamps

### LINKED ISSUE #345: Allow TimeSeries Cache to read same time stamps
Currently the time series cache detects reading data at the same time stamp as "arbitrary access". In the original scope we didn't expect to have to read the same time-stamp twice but for the developments in PowerSimulationsDecomposition.jl this will be a necessary feature.

daniel-thom:
Fixes #345


jd-lara (approved):

jd-lara:
@daniel-thom looks good to me. Merge whenever

PR #348, opened 2024-04-18
Store time series metadata in SQLite

daniel-thom:
New features:
- Allows addition of user-defined features to time series arrays.
- Adds support for different time series resolutions.

Refactor/re-design:
- Store time series metadata in a SQLite database instead of per-component dictionaries. This allows system-wide SQL queries instead of looping across component dictionaries.
- Consolidate management of time series in TimeSeriesManager instead of individual time series storage implementations.

Features removed:
- get_time_series and get_time_series_multiple no longer support abstract types. This could be restored, but I think it’s better this way. list_time_series* methods support abstract types.


src/hdf5_time_series_storage.jl
```diff
@@ -293,54 +276,22 @@ function get_data_type(ts::TimeSeriesData)
...
-    path,
-    rows,
-    ::Type{T},
-) where {T <: Forecast}
-    data = _read_time_series_attributes_common(storage, path, rows)
-    data["interval"] = Dates.Millisecond(HDF5.read(HDF5.attributes(path)["interval"]))
-    return data
+function _read_time_series_attributes(path)
+    return Dict(
+        "type" => _read_time_series_type(path),
```
daniel-thom, line 291: We should try to remove the need to store these in HDF5 attributes. Perhaps they can be in the metadata instead. Likely out of scope for this PR.

src/system_data.jl
```diff
@@ -620,29 +461,37 @@ function transform_single_time_series!(
     horizon::Int,
     interval::Dates.Period,
 ) where {T <: DeterministicSingleTimeSeries}
+    resolutions = list_time_series_resolutions(data; time_series_type = SingleTimeSeries)
+    if length(resolutions) > 1
+        # TODO: This needs to support an alternate method where horizon is expressed as a
```
daniel-thom, line 466: Out of scope for this PR. Need an issue to track this.

src/system_data.jl
```diff
@@ -972,17 +832,30 @@ function get_masked_component(data::SystemData, uuid::Base.UUID)
...
+
+list_time_series_resolutions(
+    data::SystemData;
+    time_series_type::Union{Type{<:TimeSeriesData}, Nothing} = nothing,
+) = list_time_series_resolutions(
+    data.time_series_manager.metadata_store;
+    time_series_type = time_series_type,
+)
+
+# TODO: do we need this? The old way of calculating this required a single resolution.
```
daniel-thom, line 853: If we want this feature, it will have to be implemented in a different way.

daniel-thom:
Closing because CodeCov is not currently working on forks.

PR #357, opened 2024-04-29
Refactor storage of supplemental attributes and time series

### LINKED ISSUE #360: Just a reminder that this needs to happen

daniel-thom:
- Remove supplemental_attributes_container and time_series_container from component structs and make them implement `supports_supplemental_attributes` and `supports_time_series` instead.
- Assign a `SharedSystemReferences` struct to each component via `InfrastructureSystemsInternal` so that components and attributes have access to time series data.
- Store component / supplemental attribute associations in SQLite instead of per-component dictionaries.

closes #360


src/component.jl
```diff
@@ -125,96 +115,40 @@ function assign_new_uuid_internal!(component::InfrastructureSystemsComponent)
...
-    attribute_container[T][uuid] = attribute
-    @debug "SupplementalAttribute type $T with UUID $uuid) stored in component $(summary(component))" _group =
-        LOG_GROUP_SYSTEM
-    return
-end
-
 """
 Return true if the component has supplemental attributes of the given type.
 """
 function has_supplemental_attributes(
```
jd-lara, line 121: Why change in the order of operations.

src/supplemental_attribute.jl
jd-lara, line 1: What's the change in this file? Where did the content go?

src/supplemental_attribute_associations.jl
```diff
@@ -0,0 +1,401 @@
...
+"""
+function to_records(associations::SupplementalAttributeAssociations)
+    query = "SELECT * FROM $SUPPLEMENTAL_ATTRIBUTE_TABLE_NAME"
+    return Tables.rowtable(_execute(associations, query))
+end
+
+"""
+Add records to the database. Expects output from [`to_records`](@ref).
+"""
+function load_records!(associations::SupplementalAttributeAssociations, records)
```
jd-lara, line 314: What's a record in this function?

src/supplemental_attributes.jl
```diff
@@ -1,76 +1,109 @@
...
-        throw(ArgumentError("add_supplemental_attribute! only accepts concrete types"))
-    end
+)
+    is_attached(attribute, mgr) && return
 
-    supplemental_attribute_uuid = get_uuid(supplemental_attribute)
-    if isempty(get_component_uuids(supplemental_attribute))
+    if !allow_existing_time_series && has_time_series(attribute)
         throw(
             ArgumentError(
```
jd-lara, line 65: Should we still show the supplemental attribute type in the error?

jd-lara (approved):

jd-lara:
The only minor concern I have is the use of `list_` in some cases but mostly that's for interfaces that users won't be exposed to.

src/component.jl
daniel-thom: I think I did that for consistency with `has_time_series`. I really dislike this ordering question and wish we had gone our own way. I'm not opposed to changing it. However, my thought is that it is more intuitive to put the component before the type, similar to `haskey(Dict, my_key)`.
```
"""
Return true if the component or supplemental attribute has time series data of type T.
"""
function has_time_series(
    val::TimeSeriesOwners,
    ::Type{T},
) where {T <: TimeSeriesData}
```

src/supplemental_attribute.jl
daniel-thom: Attributes no longer store component UUIDs, so we don't need any of it.

src/supplemental_attribute_associations.jl
daniel-thom: Rows from the database. When we serialize the system to JSON we write all database rows (`to_records`) to the JSON file. When we de-serialize from JSON we load those records back into the database.

src/supplemental_attributes.jl
daniel-thom: Yeah. I'll change this to call `summary` on the attribute.

src/utils/sqlite.jl
```diff
@@ -37,3 +37,22 @@ function backup(
...
+"""
+Wrapper around SQLite.DBInterface.execute to provide debug log messages.
+"""
+function execute(db::SQLite.DB, query::AbstractString, log_group::Symbol)
+    @debug "Execute SQL" _group = log_group query
+    return SQLite.DBInterface.execute(db, query)
+end
+
+"""
+Run a query to find a count. The query must produce a column called count with one row.
```
GabrielKS, line 50: Current behavior seems to ignore rows 2 and up, should this be an error instead?

src/serialization.jl
```diff
@@ -210,7 +210,7 @@ function deserialize(::Type{T}, data::Dict) where {T <: NamedTuple}
 end
 
 # Some types that definitely won't be deserialized from Dicts
-const _NOT_FROM_DICT = Union{Nothing, Real, String}
+const _NOT_FROM_DICT = Union{Nothing, Real, AbstractString}
```
GabrielKS, line 213: thanks :)

src/time_series_interface.jl
```diff
@@ -500,6 +491,28 @@ function clear_time_series!(owner::TimeSeriesOwners)
     return
 end
 
+"""
+This function must be called when a component or attribute is removed from a system.
+"""
+function prepare_for_removal!(owner::TimeSeriesOwners)
+    clear_time_series!(owner)
+    set_shared_system_references!(owner, nothing)
```
GabrielKS, line 499: Why is this necessary?

GabrielKS (commented):
A very quick review, no guarantees of thoroughness.

src/component.jl
daniel-thom: Both now exist.

src/component.jl
```diff
@@ -125,96 +115,40 @@ function assign_new_uuid_internal!(component::InfrastructureSystemsComponent)
...
 
 function clear_supplemental_attributes!(component::InfrastructureSystemsComponent)
-    container = get_supplemental_attributes_container(component)
-    for attributes in values(container)
-        for attribute in collect(values(attributes))
-            detach_component!(attribute, component)
-            detach_supplemental_attribute!(component, attribute)
-        end
+    mgr = _get_supplemental_attributes_manager(component)
+    isnothing(mgr) && return
```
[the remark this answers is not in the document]
daniel-thom: A component might not support supplemental attributes.

src/component.jl
```diff
@@ -228,27 +162,94 @@ function get_supplemental_attributes(
...
 ) where {T <: SupplementalAttribute}
-    return get_supplemental_attributes(
-        filter_func,
-        T,
-        get_supplemental_attributes_container(component),
+    return _get_supplemental_attributes(filter_func, T, component)
+end
+
+function get_supplemental_attributes(
+    filter_func::Function,
```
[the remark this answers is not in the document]
daniel-thom: Per our conversation, I'm leaving this alone for now.

src/internal.jl
```diff
@@ -13,26 +13,37 @@ end
 serialize(val::SystemUnitsSettings) = serialize_struct(val)
 deserialize(T::Type{<:SystemUnitsSettings}, val::Dict) = deserialize_struct(T, val)
 
+@kwdef mutable struct SharedSystemReferences <: InfrastructureSystemsType
```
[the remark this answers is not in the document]
daniel-thom: Fair point. I originally wanted to include units_info which could be mutated, I think. But nothing is being mutated now, so I'll make it immutable.

src/internal.jl
```diff
@@ -13,26 +13,37 @@ end
...
+    time_series_manager::Any = nothing
+end
+
 """
 Internal storage common to InfrastructureSystems types.
 """
 mutable struct InfrastructureSystemsInternal <: InfrastructureSystemsType
     uuid::Base.UUID
+    shared_system_references::Union{Nothing, SharedSystemReferences}
     units_info::Union{Nothing, UnitsData}
```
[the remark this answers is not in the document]
daniel-thom: Yes, I wanted that, too. However, (1) it's getting set from PSY and (2) it needs to be set to something valid before the component is part of the system. I decided to defer considering this point until later.

src/internal.jl
```diff
@@ -13,26 +13,37 @@ end
...
+    ext = nothing,
+) =
+    InfrastructureSystemsInternal(uuid, shared_system_references, units_info, ext)
 
 """
 Creates InfrastructureSystemsInternal with an existing UUID.
 """
-InfrastructureSystemsInternal(u::UUIDs.UUID) =
-    InfrastructureSystemsInternal(u, nothing, nothing)
+InfrastructureSystemsInternal(u::Base.UUID) =
```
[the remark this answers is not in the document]
daniel-thom: I changed it because it made the Julia language server happy (it was complaining, which must be a bug). Ultimately, it doesn't matter.
```
julia> UUIDs.UUID === Base.UUID
true
```

src/time_series_interface.jl
daniel-thom: To make sure that a removed component doesn't have any references from the system.

src/supplemental_attribute_associations.jl
```diff
@@ -0,0 +1,401 @@
...
+    LOG_GROUP_SUPPLEMENTAL_ATTRIBUTES
+    return
+end
+
+"""
+Drop the supplemental attribute associations table.
+"""
+function drop_table(associations::SupplementalAttributeAssociations)
+    _execute(associations, "DROP TABLE IF EXISTS $SUPPLEMENTAL_ATTRIBUTE_TABLE_NAME")
+    @debug "Dropped the table $SUPPLEMENTAL_ATTRIBUTE_TABLE_NAME" _group =
```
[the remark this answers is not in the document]
daniel-thom: We don't really care because we will immediately rebuild the table. I'll make the message more generic.

src/utils/sqlite.jl
daniel-thom: We had some related conversation in the last PR, but sure, we can assert this. The use case is that the user has has asked for a count and there can only be one row.

daniel-thom: Nevermind. The SQLite.jl people don't seem to have fully implemented the Tables interface.
```
MethodError: no method matching length(::SQLite.Query{false})
```
I'm confident that this is not a problem.

GabrielKS: Okay

PR #362, opened 2024-05-01
Fix `compare_values` for arrays with `NaN`, add deserialization for more union types

GabrielKS:
This solves two issues:

1. Now that `PiecewiseStepData` HDF-serializes to an array that includes `NaN`, we need to allow `NaN` to be equal to itself when checking whether two things are the same — this is accomplished by changing `==` to `isequal`.
2. The new PowerSystems `ThermalGenerationCost` has a field of type `Union{StartUpStages, Float64}`, where `StartUpStages` is a `NamedTuple`. We didn't have code to deserialize that, but we did have code to deserialize `Union{NamedTuple, Nothing}`, so I generalized it.

I would have liked to have done most of the logic involved in (2) using the type system, but I did not seem to be able to do so without angering `detect_unbound_args`. For instance, this method solves the problem (though less generally) but is said to have unbound type parameters:
```julia
deserialize(::Type{Union{A, B}}, data::Dict) where {A <: _NOT_FROM_DICT, B <: NamedTuple} = deserialize(B, data)
```
I'm not sure I believe `detect_unbound_args` that there are unbound type parameters here, but I have opted for a more manual, less elegant, possibly more flexible approach.

### DIFF
--- a/src/serialization.jl
+++ b/src/serialization.jl
@@ -209,11 +209,22 @@ function deserialize(::Type{T}, data::Dict) where {T <: NamedTuple}
     return T(key = data[string(key)] for key in fieldnames(T))
 end
 
-function deserialize(
-    ::Type{T},
-    data::Union{Nothing, Dict},
-) where {T <: Union{Nothing, NamedTuple}}
-    return isnothing(data) ? nothing : deserialize(T.b, data)
+# Some types that definitely won't be deserialized from Dicts
+const _NOT_FROM_DICT = Union{Nothing, Real, String}
+
+# If deserializing into a Union of some _NOT_FROM_DICT and something else (e.g., a
+# NamedTuple) and we are given a Dict as input data, pick the something else. NOTE: it would
+# be great to do this purely using the type system. I found ways to do subsets of the task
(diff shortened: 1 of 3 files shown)


src/serialization.jl
```diff
@@ -209,11 +209,22 @@ function deserialize(::Type{T}, data::Dict) where {T <: NamedTuple}
     return T(key = data[string(key)] for key in fieldnames(T))
 end
 
-function deserialize(
-    ::Type{T},
-    data::Union{Nothing, Dict},
-) where {T <: Union{Nothing, NamedTuple}}
-    return isnothing(data) ? nothing : deserialize(T.b, data)
+# Some types that definitely won't be deserialized from Dicts
+const _NOT_FROM_DICT = Union{Nothing, Real, String}
```
daniel-thom, line 213: I'm not sure if it's ever possible to get another type of string, but we might want to use`AbstractString` instead.

daniel-thom (approved):

jd-lara (approved):

jd-lara (approved):

PR #364, opened 2024-05-05
Fix handling of deepcopy for SQLite databases

daniel-thom:
This PR fixes a bug introduced by #357. The code did not properly handle the shared references to the internal SQLite database during a deepcopy of the system. This fix does the following:
- Use separate SQLite in-memory databases for time series metadata and supplemental attribute associations.
- Implemental `deepcopy_internal` for structs that do not support the Julia default method at the lowest possible level.

An alternate is solution is to implement management of the SQLite database at the `SystemData` level. I did not go that route because of the requirement that supplemental attributes be persisted to JSON, not SQLite, during serialization.


jd-lara (approved):
The notes about the design seem fine to me.

PR #365, opened 2024-05-08
Redefine TimeSeriesKey

daniel-thom:
Change the TimeSeriesInfo structs to be keys and return them from add_time_series!(). Remove the old TimeSeriesKey.

### DIFF
--- a/src/system_data.jl
+++ b/src/system_data.jl
@@ -269,7 +269,7 @@ Removes all time series of a particular type from a System.
 function remove_time_series!(data::SystemData, ::Type{T}) where {T <: TimeSeriesData}
     _throw_if_read_only(data.time_series_manager)
     for component in iterate_components_with_time_series(data; time_series_type = T)
-        for ts_metadata in list_time_series_metadata(component; time_series_type = T)
+        for ts_metadata in get_time_series_metadata(component; time_series_type = T)
             remove_time_series!(data, component, ts_metadata)
         end
     end
@@ -466,7 +466,7 @@ function transform_single_time_series!(
     horizon::Int,
     interval::Dates.Period,
 ) where {T <: DeterministicSingleTimeSeries}
-    resolutions = list_time_series_resolutions(data; time_series_type = SingleTimeSeries)
+    resolutions = get_time_series_resolutions(data; time_series_type = SingleTimeSeries)
     if length(resolutions) > 1
         # TODO: This needs to support an alternate method where horizon is expressed as a
         # Period (horizon * resolution)
@@ -869,10 +869,10 @@ get_forecast_initial_timestamp(data::SystemData) =
 get_forecast_interval(data::SystemData) =
     get_forecast_interval(data.time_series_manager.metadata_store)
 
-list_time_series_resolutions(
+get_time_series_resolutions(
     data::SystemData;
     time_series_type::Union{Type{<:TimeSeriesData}, Nothing} = nothing,
-) = list_time_series_resolutions(
+) = get_time_series_resolutions(
     data.time_series_manager.metadata_store;
     time_series_type = time_series_type,
 )
--- a/src/system_data.jl
+++ b/src/system_data.jl
@@ -149,15 +149,17 @@ function add_time_series_from_file_metadata!(
     file_metadata::Vector{TimeSeriesFileMetadata};
     resolution = nothing,
 ) where {T <: InfrastructureSystemsComponent}
+    ts_keys = Vector{TimeSeriesKey}(undef, length(file_metadata))
     open_time_series_store!(data, "r+") do
         cache = TimeSeriesParsingCache()
-        for metadata in file_metadata
+        for (i, metadata) in enumerate(file_metadata)
             if resolution === nothing || metadata.resolution == resolution
-                add_time_series_from_file_metadata_internal!(data, T, cache, metadata)
+                ts_keys[i] =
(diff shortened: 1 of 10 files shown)


jd-lara (approved):

src/time_series_metadata_store.jl
```diff
@@ -674,8 +674,8 @@ end
 Return information about each time series array attached to the owner.
 This information can be used to call get_time_series.
 """
-function list_time_series_info(store::TimeSeriesMetadataStore, owner::TimeSeriesOwners)
-    return [make_time_series_info(x) for x in list_metadata(store, owner)]
+function list_time_series_keys(store::TimeSeriesMetadataStore, owner::TimeSeriesOwners)
```
jd-lara, line 677:
```suggestion
function get_time_series_keys(store::TimeSeriesMetadataStore, owner::TimeSeriesOwners)
```

src/system_data.jl
```diff
@@ -208,14 +210,19 @@ function add_time_series!(
     time_series::TimeSeriesData;
     features...,
 )
+    key = nothing
     for component in components
-        add_time_series!(
+        # Component information is not embedded into the key and so it will always be the
+        # same.
+        key = add_time_series!(
```
GabrielKS, line 217: For a bit more robustness we could keep track of the previous key and assert that they're equal — might not be necessary

src/utils/print.jl
```diff
@@ -321,19 +322,20 @@ end
...
+        if !haskey(data_by_type, ts_type)
+            data_by_type[ts_type] = Vector{OrderedDict{String, Any}}()
         end
         data = OrderedDict{String, Any}()
-        for field in fieldnames(typeof(info))
-            if field == :type
-                data[string(field)] = string(nameof(Base.getproperty(info, field)))
+        for (fname, ftype) in zip(fieldnames(typeof(key)), fieldtypes(typeof(key)))
+            if ftype <: TimeSeriesData
+                data[string(fname)] = string(nameof(Base.getproperty(key, fname)))
```
GabrielKS, line 333: Noting that this isn't covered by tests

test/test_time_series.jl
```diff
@@ -2582,3 +2589,17 @@ end
...
+@testset "Test serialization of time series keys" begin
+    key = IS.StaticTimeSeriesKey(
+        IS.SingleTimeSeries,
+        "test",
+        Dates.now(),
+        Dates.Hour(1),
+        12,
+        Dict("scenario" => "high"),
+    )
+    data = IS.serialize(key)
```
GabrielKS, line 2602: Next line implicitly does this on the current implementation, but maybe test that data isa dictionary?

GabrielKS (changes requested):
Mostly looks good, a few minor additions requested

src/system_data.jl
daniel-thom: I wanted to do that, but you can't compare immutable structs with Dicts automatically...or at least I didn't find an easy way in a short search. I'm very confident that it is ok.

src/utils/print.jl
daniel-thom: I think it is actually incorrect.

src/system_data.jl
GabrielKS: Okay

src/time_series_metadata_store.jl
daniel-thom: This is unfortunate. I think that get vs list adds clarity and intuition for users. Using list will inform people that they don't need to `collect`. But I will relent.

test/test_time_series.jl
daniel-thom: I will change it so that it is very obvious that the test doesn't care about what `data` is. :)

GabrielKS: I think the test _should_ care about what `data` is, `data` needs to be something that is trivially serializable, like a `dict` or a primitive. At a certain point `IS.serialize` falls back to just returning the struct itself and that should be tested against.

GabrielKS: Or the test should do a round trip through JSON writing and reading.

daniel-thom: All I want to do is verify that a round trip works. If someone changes the intermediate type, I don't want to have to update the test. Fair point about the default fallback functions for serialize and deserialize. That can be solved by `@test key2 !== key1`. But, that actually doesn't work.

I think our fallback functions are unsafe. They should be returning a deepcopy of the inputs. Someone would be very surprised if they called serialize on a struct that contains a Dict, changed the returned value, and then observed that the original Dict was also changed. This will add some minor cost because of copies, but I think we should do it. It's unlikely but possible that someone could do this with an ext field.

GabrielKS: I agree that the fallbacks are unsafe. But even if they were better, I think we need to be verifying that the result of `serialize` is something that can actually be preserved under JSON serialization. To avoid relying on the intermediate type being dict, what about something like
```julia
intermediate_serialized = IS.serialize(key)
key2 = IS.deserialize(typeof(key), JSON3.read(JSON3.write(intermediate_serialized), typeof(intermediate_serialized)))
```

? If we find a good general solution, it should be encapsulated somehow and I'll use it [here](https://github.com/NREL-Sienna/InfrastructureSystems.jl/blob/508f4cf3a71cfade8c46f32c19c3d3ca40b9380b/test/test_function_data.jl#L120) and I'm sure there are other places.

GabrielKS (approved):

PR #367, opened 2024-05-09
Accommodate `NaN` in `FunctionData` validation and comparison, extend serialization to more types

GabrielKS:
The market bid cost curve representation relies on putting `NaN`s in `PiecewiseStepData`, so I took the opportunity to make `==`, `isequal`, and `hash` work nicely for `FunctionData` in general.

### DIFF
--- a/src/function_data.jl
+++ b/src/function_data.jl
@@ -88,8 +88,10 @@ end
 function _validate_piecewise_x(x_coords)
     (length(x_coords) < 2) &&
         throw(ArgumentError("Must specify at least two x-coordinates"))
-    !issorted(x_coords) &&
+    # This could be generalized to allow NaNs in more places
+    if !(issorted(x_coords) || (isnan(first(x_coords)) && issorted(x_coords[2:end])))
         throw(ArgumentError("Piecewise x-coordinates must be ascending, got $x_coords"))
+    end
 end
 
 """
@@ -285,12 +287,11 @@ Base.length(pwl::Union{PiecewiseLinearData, PiecewiseStepData}) =
 Base.getindex(pwl::PiecewiseLinearData, ix::Int) =
     getindex(get_points(pwl), ix)
 
-Base.:(==)(a::PiecewiseLinearData, b::PiecewiseLinearData) =
-    get_points(a) == get_points(b)
+Base.:(==)(a::T, b::T) where T <: FunctionData = double_equals_from_fields(a, b)
 
-Base.:(==)(a::PiecewiseStepData, b::PiecewiseStepData) =
-    (get_x_coords(a) == get_x_coords(b)) &&
-    (get_y_coords(a) == get_y_coords(b))
+Base.isequal(a::T, b::T) where T <: FunctionData = isequal_from_fields(a, b)
+
+Base.hash(a::FunctionData) = hash_from_fields(a)
 
 function _slope_convexity_check(slopes::Vector{Float64})
     for ix in 1:(length(slopes) - 1)
--- a/src/function_data.jl
+++ b/src/function_data.jl
@@ -85,7 +85,7 @@ function transform_array_for_hdf(
     return transform_array_for_hdf(transfd_data)
 end
 
-function _validate_piecewise_x(x_coords)
+function _validate_piecewise_x(x_coords::Vector)
     (length(x_coords) < 2) &&
         throw(ArgumentError("Must specify at least two x-coordinates"))
     # This could be generalized to allow NaNs in more places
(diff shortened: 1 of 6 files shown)


src/function_data.jl
```diff
@@ -88,8 +88,10 @@ end
 function _validate_piecewise_x(x_coords)
```
jd-lara, line 88: should `x_coords` be typed?

src/utils/utils.jl
```diff
@@ -595,3 +595,31 @@ function compute_sha256(filename::AbstractString)
...
+
+"Compute the conjunction of the `isequal` values of all the fields in `a` and `b`"
+isequal_from_fields(a::T, b::T) where {T} =
+    compare_over_fields(isequal, &, true, a, b)
+
+"Compute a hash of the instance `a` by combining hashes of all its fields"
+hash_from_fields(a) = hash_from_fields(a, zero(UInt))
+
+function hash_from_fields(a, h::UInt)
+    for field in sort(collect(fieldnames(typeof(a))))
```
daniel-thom, line 621: Nit: sort in place with `sort!`

src/function_data.jl
```diff
@@ -285,12 +287,11 @@ Base.length(pwl::Union{PiecewiseLinearData, PiecewiseStepData}) =
 Base.getindex(pwl::PiecewiseLinearData, ix::Int) =
     getindex(get_points(pwl), ix)
 
-Base.:(==)(a::PiecewiseLinearData, b::PiecewiseLinearData) =
-    get_points(a) == get_points(b)
+Base.:(==)(a::T, b::T) where {T <: FunctionData} = double_equals_from_fields(a, b)
```
daniel-thom, line 290: The implementation is good. Are there any differences with https://github.com/JuliaServices/AutoHashEquals.jl? If they are equivalent, any reason not to use that package? I'm not opposed to having this here.

You might want to mention Julia issue 4648. Seems like the default behavior may change in the future.

test/test_function_data.jl
```diff
@@ -143,3 +159,20 @@ end
...
+
+    # Value-equal function data should be == except when containing NaN since NaN != NaN;
+    # value-equal function data should be isequal even when containing NaN; hash equality
+    # should correspond with isequal
+    for my_type in IS.get_all_concrete_subtypes(IS.FunctionData)
+        @test examples_1[(my_type, false)] == examples_2[(my_type, false)]
+        @test examples_1[(my_type, true)] != examples_2[(my_type, true)]
+        @test isequal(examples_1[(my_type, false)], examples_2[(my_type, false)])
+        @test isequal(examples_1[(my_type, true)], examples_2[(my_type, true)])
+        @test hash(examples_1[(my_type, false)]) == hash(examples_2[(my_type, false)])
```
daniel-thom, line 175: Negative tests for isequal and hash?

src/serialization.jl
```diff
@@ -209,8 +209,8 @@ function deserialize(::Type{T}, data::Dict) where {T <: NamedTuple}
     return T(key = data[string(key)] for key in fieldnames(T))
 end
 
-# Some types that definitely won't be deserialized from Dicts
-const _NOT_FROM_DICT = Union{Nothing, Real, AbstractString}
+# Some types that definitely won't be deserialized from raw Dicts
+const _NOT_FROM_DICT = Union{Nothing, Real, AbstractString, TimeSeriesKey}
```
daniel-thom, line 213: I'm puzzled by this change. Can you explain?

GabrielKS: The deserialization code first checks to see if the dictionary to be deserialized contains concrete type information. If it does, it will deserialize to that directly and we'll never get to this union method. `TimeSeriesKey`s are always serialized with time series information (as are most structs, this list could be much longer) and therefore they will never be the answer when we get to this method.

src/function_data.jl
GabrielKS: I think AutoHashEquals should work — @jd-lara okay to add as a dependency?

src/function_data.jl
GabrielKS: Done

src/utils/utils.jl
GabrielKS: Done

test/test_function_data.jl
GabrielKS: Done

src/function_data.jl
jd-lara: If don't need it and this works we should keep our own method instead of more dependencies.

jd-lara (approved):

PR #374, opened 2024-06-04
Add default value to struct docstrings

kdayday:
Adds a print of the default value of parameters in each parameter docstring in auto-generated structs

Example from `ThermalStandard`
Screenshot 2024-06-04 at 2 10 58 PM

### DIFF
--- a/src/utils/generate_structs.jl
+++ b/src/utils/generate_structs.jl
@@ -19,7 +19,7 @@ This file is auto-generated. Do not edit.
 
 # Arguments
 {{#parameters}}
-- `{{name}}::{{{data_type}}}`:{{#default}} (default: `{{{default}}}`){{/default}}{{#comment}} {{{comment}}}{{/comment}}{{#valid_range}}, validation range: `{{{valid_range}}}`{{/valid_range}}{{#validation_action}}, action if invalid: `{{{validation_action}}}`{{/validation_action}}
+- `{{name}}::{{{data_type}}}`:{{#default}} (default: `{{{default}}}`){{/default}}{{#comment}} {{{comment}}}{{/comment}}{{#valid_range}}, validation range: `{{{valid_range}}}`{{/valid_range}}
 {{/parameters}}
 \"\"\"
 mutable struct {{struct_name}}{{#parametric}}{T <: {{parametric}}}{{/parametric}} <: {{supertype}}
--- a/src/utils/generate_structs.jl
+++ b/src/utils/generate_structs.jl
@@ -19,7 +19,7 @@ This file is auto-generated. Do not edit.
 
 # Arguments
 {{#parameters}}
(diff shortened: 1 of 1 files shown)


jd-lara:
@kdayday we can populate the `(optional)` here with (has default) for the fields that have defaults and autodetect here in the template instead of doing it manually.

src/utils/generate_structs.jl
```diff
@@ -19,7 +19,7 @@ This file is auto-generated. Do not edit.
 
 # Arguments
 {{#parameters}}
-- `{{name}}::{{{data_type}}}`{{#comment}}: {{{comment}}}{{/comment}}{{#valid_range}}, validation range: `{{{valid_range}}}`{{/valid_range}}{{#validation_action}}, action if invalid: `{{{validation_action}}}`{{/validation_action}}
+- `{{name}}::{{{data_type}}}`{{#comment}}: {{{comment}}}{{/comment}}{{#default}}, default value: `{{{default}}}`{{/default}}{{#valid_range}}, validation range: `{{{valid_range}}}`{{/valid_range}}{{#validation_action}}, action if invalid: `{{{validation_action}}}`{{/validation_action}}
```
daniel-thom, line 22: Can we do something slightly different with `, default value`? It looks weird when the comment ends with a period. My original suggestion was to put this just before the comment.

kdayday:
> @kdayday we can populate the `(optional)` here with (has default) for the fields that have defaults and autodetect here in the template instead of doing it manually.

@jd-lara I'm going to delete the `(optional)`s on PSY now. Is that what you meant? We'll just have the default value then

jd-lara:
> > @kdayday we can populate the `(optional)` here with (has default) for the fields that have defaults and autodetect here in the template instead of doing it manually.
> 
> @jd-lara I'm going to delete the `(optional)`s on PSY now. Is that what you meant? We'll just have the default value then

I mean that we can add a note to the field that is that defaults when they are just in case the users misses the fact that there is a default printed in the long comment

src/utils/generate_structs.jl
kdayday: @jd-lara @daniel-thom  How's this, in front in parentheses?

Also removed the ending periods https://github.com/NREL-Sienna/PowerSystems.jl/pull/1111/commits/a23061991537da8536d887fed5ae169bdc76e05b 

We could also remove the `action if invalid` to shorten the strings, since the user will be able to see the warning or error anyways

Screenshot 2024-06-04 at 5 04 19 PM

daniel-thom: Looks good to me. I agree with your point about `action if invalid`. I doubt anyone relies on that.

kdayday: Removed action if invalid below

daniel-thom (approved):
Looks good

jd-lara (approved):

PR #379, opened 2024-06-13
add new method for checking if there are components of a particular type

jd-lara:


src/components.jl
```diff
@@ -200,6 +200,23 @@ function has_component(
     return haskey(components.data[T], name)
 end
 
+"""
+Check to see if a component exists.
```
daniel-thom, line 204:
```suggestion
Check to see if a component if the given type exists.
```

daniel-thom (approved):

PR #380, opened 2024-06-18
Time series function documentation

kdayday:
Add/copy docs strings for get_time_series and get_time_series_keys so they can be imported/exported by PSY. Resulting docs over in PSY:

Screenshot 2024-06-18 at 11 32 25 AM
Screenshot 2024-06-18 at 11 32 41 AM

### DIFF
--- a/src/time_series_interface.jl
+++ b/src/time_series_interface.jl
@@ -51,6 +51,21 @@ function get_time_series(
     return deserialize_time_series(T, storage, ts_metadata, rows, columns)
 end
 
+"""
+Return a time series corresponding to the given parameters.
+
+# Arguments
+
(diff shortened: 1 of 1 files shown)


daniel-thom (approved):

jd-lara (commented):
@daniel-thom can you confirm if we already have in the IS documentation the bulk addition of time series? If not, should we add it?

daniel-thom:
> @daniel-thom can you confirm if we already have in the IS documentation the bulk addition of time series? If not, should we add it?

This [section](https://github.com/NREL-Sienna/PowerSystems.jl/blob/psy4/docs/src/modeler_guide/time_series.md?plain=1#L253) is in PowerSystems. This is about keeping the HDF5 file open, not adding many arrays at once.

PR #381, opened 2024-06-21
Optimize performance of adding many time series at once

daniel-thom:
- Provide a procedure to add many time series at once.
- Support the case of `transform_single_time_series!`.
- Use SQL parameters in order to protect against SQL injection attacks.
- Move time series checks to the constructors instead of at the time of `add_time_series!`.
- Add debug timings to the code base with TimerOutputs.jl.


src/supplemental_attribute_associations.jl
```diff
@@ -207,23 +209,25 @@ function has_association(
     query = """
         SELECT attribute_uuid
         FROM $SUPPLEMENTAL_ATTRIBUTE_TABLE_NAME
-        WHERE attribute_uuid = '$a_uuid' AND component_uuid = '$c_uuid'
+        WHERE attribute_uuid = ? AND component_uuid = ?
```
jd-lara, line 212:
```suggestion
        WHERE attribute_uuid = ? AND component_uuid = ?
```
These are left as question marks on purpose?

daniel-thom: It is not recommended to insert user-defined strings in SQL queries because they could be SQL injection attacks. That is extremely unlikely in our case because the users would be attacking themselves. However, we should still follow best practices because it is theoretically possible that we someday use a shared database.

jd-lara (approved):

src/supplemental_attribute_associations.jl
jd-lara: nvm about that request. This is all over the code base.

PR #387, opened 2024-07-19
Add option to specify equality predicate for `compare_values` and provide a more sensible default (IS)

GabrielKS:
Previously, at the leaves of `compare_values`'s recursion tree, it would mostly test equality using the `==` operator, sometimes the `Base.isequal` function. This PR lets a sophisticated user specify a custom equality predicate `match_fn` to override that behavior, such as to test a more approximate equality on a system when it is exported to a certain format and reimported. It also standardizes the default `match_fn` to
```julia
isequivalent(x, y) = isequal(x, y) || (x == y)
```
, which is `true` for `NaN, NaN` (unlike `==`) and for `-0.0, 0.0` (unlike `isequal`). Is that a violation of SemVer?

### DIFF
--- a/src/time_series_metadata_store.jl
+++ b/src/time_series_metadata_store.jl
@@ -1044,6 +1044,7 @@ function compare_values(
     y::TimeSeriesMetadataStore;
     compare_uuids = false,
     exclude = Set{Symbol}(),
+    match_fn = isequivalent,
 )
     # Note that we can't compare missing values.
     owner_uuid = compare_uuids ? ", owner_uuid" : ""
@@ -1053,7 +1054,7 @@ function compare_values(
     """
     table_x = Tables.rowtable(_execute(x, query))
     table_y = Tables.rowtable(_execute(y, query))
-    return table_x == table_y
+    return match_fn(table_x, table_y)
 end
 
 ### Non-TimeSeriesMetadataStore functions ###
--- a/src/time_series_metadata_store.jl
+++ b/src/time_series_metadata_store.jl
@@ -1040,11 +1040,11 @@ function _try_time_series_metadata_by_full_params(
 end
 
 function compare_values(
+    match_fn::Union{Function, Nothing},
     x::TimeSeriesMetadataStore,
     y::TimeSeriesMetadataStore;
     compare_uuids = false,
     exclude = Set{Symbol}(),
-    match_fn = isequivalent,
 )
     # Note that we can't compare missing values.
     owner_uuid = compare_uuids ? ", owner_uuid" : ""
@@ -1054,6 +1054,7 @@ function compare_values(
     """
     table_x = Tables.rowtable(_execute(x, query))
     table_y = Tables.rowtable(_execute(y, query))
+    match_fn = _fetch_match_fn(match_fn)
     return match_fn(table_x, table_y)
 end
 
--- a/src/utils/utils.jl
+++ b/src/utils/utils.jl
@@ -131,11 +131,12 @@ function compare_values(
     y::T;
     compare_uuids = false,
     exclude = Set{Symbol}(),
+    match_fn = isequivalent,
 ) where {T}
     match = true
     fields = fieldnames(T)
     if isempty(fields)
-        match = x == y
+        match = match_fn(x, y)
     else
         for field_name in fields
             field_name in exclude && continue
@@ -147,6 +148,7 @@ function compare_values(
                     val2;
                     compare_uuids = compare_uuids,
                     exclude = exclude,
+                    match_fn = match_fn,
                 )
                     @error "values do not match" T field_name val1 val2
                     match = false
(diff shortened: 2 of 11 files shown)


jd-lara (commented):
I am ok with this PR but I don't understand why is breaking PSY. We need to make sure this change is backwards compatible.

GabrielKS:
Okay, the interface should be backwards compatible now.

src/utils/utils.jl
```diff
@@ -115,57 +116,77 @@ function validate_exported_names(mod::Module)
...
+function compare_values(match_fn::Union{Function, Nothing}, x::T, y::U;
+    compare_uuids = false, exclude = Set{Symbol}()) where {T, U}
+    # Special case: if match_fn is nothing, try calling the two-argument version to maintain
+    # backwards compatibility with packages that only implement that. Keep track of this to
+    # avoid infinite recursion. TODO remove in next major version
+    if isnothing(match_fn) && !(COMPARE_VALUES_SENTINEL in exclude)
+        return compare_values(x, y; compare_uuids = compare_uuids,
+            exclude = union(exclude, [COMPARE_VALUES_SENTINEL]))
+    end
+    exclude = setdiff(exclude, [COMPARE_VALUES_SENTINEL])
```
GabrielKS, line 153: This is a little hacky but I am convinced it is the least bad way to accomplish the task without changing the interface at all (and it's temporary)

GabrielKS: (ask me what I learned about `Core.kwcall` today)

jd-lara: lol ...

src/time_series_metadata_store.jl
```diff
@@ -1040,6 +1040,7 @@ function _try_time_series_metadata_by_full_params(
 end
 
 function compare_values(
+    match_fn::Union{Function, Nothing},
```
jd-lara, line 1043: Wait, what is supposed to happen here is a user passes a nothing?

src/utils/utils.jl
```diff
@@ -115,57 +116,77 @@ function validate_exported_names(mod::Module)
     return is_valid
 end
 
+# Make match_fn optional
+compare_values(x, y; kwargs...) = compare_values(nothing, x, y; kwargs...)
+
+# Get the default match_fn if necessary. Only call when you know you're done recursing
+_fetch_match_fn(match_fn::Function) = match_fn
```
jd-lara, line 123: This is quite a complicated workaround... I understand it avoids code repetition but it makes the code harder to read

jd-lara (changes requested):
The PR looks good to me as it is. But I need you to document this function and its use somewhere in the docs because it isn't obvious what to use it for.

src/time_series_metadata_store.jl
GabrielKS: We can't just do
```julia
compare_values(x, y; kwargs...) = compare_values(isequivalent, x, y; kwargs...)
```
because we need to be able to keep track of the fact that the two-argument version was initially called so that an old two-argument version in PSY can be called (this was the backwards compatibility issue with my first attempt). So we do
```julia
compare_values(x, y; kwargs...) = compare_values(nothing, x, y; kwargs...)
```
and define that `nothing` will eventually default to `isequivalent` if it isn't intercepted by a two-argument version. This is documented here:
https://github.com/NREL-Sienna/InfrastructureSystems.jl/blob/295e3656f948c80f6d3bcfb10e9b1788e1553c0c/src/utils/utils.jl#L136

src/utils/utils.jl
GabrielKS: Would
```julia
(isnothing(match_fn) ? isequivalent : match_fn)(a, b)
```
be more readable than
```julia
_fetch_match_fn(match_fn)(a, b)
```
?

GabrielKS:
> The PR looks good to me as it is. But I need you to document this function and its use somewhere in the docs because it isn't obvious what to use it for.

There is a docstring at https://github.com/NREL-Sienna/InfrastructureSystems.jl/blob/295e3656f948c80f6d3bcfb10e9b1788e1553c0c/src/utils/utils.jl#L132 which will end up [here](https://nrel-sienna.github.io/InfrastructureSystems.jl/stable/InfrastructureSystems/#InfrastructureSystems.compare_values-Union%7BTuple%7BT%7D,%20Tuple%7BT,%20T%7D%7D%20where%20T), are you talking about something more than that?

src/utils/utils.jl
```diff
@@ -115,57 +116,77 @@ function validate_exported_names(mod::Module)
...
   - `exclude::Set{Symbol} = Set{Symbol}(): Fields to exclude from comparison. Passed on
      recursively and so applied per type.
 """
-function compare_values(
-    x::T,
-    y::T;
-    compare_uuids = false,
-    exclude = Set{Symbol}(),
-) where {T}
+function compare_values(match_fn::Union{Function, Nothing}, x::T, y::U;
```
daniel-thom, line 144: @GabrielKS In what cases do we pass different types to this function? How can two values that are of different types be considered equivalent?

GabrielKS: We want for instance `isequivalent(0, 0.0)`, both because it makes sense and for backwards compatibility. In a broader sense we want to impose minimal constraints on how `match_fn` works, if it wants to specify that things of different type don't match then it is free to do so.

daniel-thom: Isn't the purpose of `compare_values` to `"Recursively compares struct values."`? We should always be comparing identical types. Are we ever calling it to test integers and floats?

PR #389, opened 2024-07-27
Document function arguments and move docstring details over from PSY

kdayday:
- documents arguments for Time series Interface functions used in PSY
- PSY modeler guides also contained details and examples about Time series functions, which have been moved or duplicated to the respective functions
- add some missing docstrings (e.g., iterate_windows)

### DIFF
--- a/src/single_time_series.jl
+++ b/src/single_time_series.jl
@@ -6,12 +6,13 @@
         internal::InfrastructureSystemsInternal
     end
 
-A deterministic forecast for a particular data field in a Component.
+A single column of time series data for a particular data field in a Component.
 
 # Arguments
 
   - `name::String`: user-defined name
   - `data::TimeSeries.TimeArray`: timestamp - scalingfactor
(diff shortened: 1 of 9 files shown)


src/single_time_series.jl
```diff
@@ -6,12 +6,13 @@
         internal::InfrastructureSystemsInternal
     end
 
-A deterministic forecast for a particular data field in a Component.
+A single column of time series data for a particular data field in a Component.
```
jd-lara, line 9: We should explain this better 

```suggestion
A single column of time series data for a particular data field in a Component. This is used to represent a single realization or scenario for the time series. 
```

kdayday: How about: `In contrast with a forecast, this can represent one continual time series,
such as a series of historical measurements or realizations or a single scenario
(e.g. a weather year or different input assumptions).`

jd-lara: sounds good

kdayday: Done

Issue #391, opened 2024-08-01
Be More Intentional About Generated Exports


GabrielKS:
It was a surprise to @daniel-thom and me that InfrastructureSystems actually exports some things:
https://github.com/NREL-Sienna/InfrastructureSystems.jl/blob/a0fe5964ad9bdbbb21df37a506df4d9731aeff4c/src/generated/includes.jl#L6
This comes from auto-generated code. We should figure out to what extent this is actually what we want.

See also https://github.com/NREL-Sienna/InfrastructureSystems.jl/issues/388

PR #392, opened 2024-08-02
Move `ValueCurve`s, cost aliases, `CostCurve`, `FuelCurve`, and associated tests from PSY to IS (IS version)

GabrielKS:
These features are now needed in packages that don't have a dependency on PowerSystems, so they need to be moved to InfrastructureSystems. The PRs are pretty straightforward, almost entirely just copypasting.


src/InfrastructureSystems.jl
```diff
@@ -2,6 +2,10 @@ isdefined(Base, :__precompile__) && __precompile__()
 
 module InfrastructureSystems
 
+# Cost aliases don't display properly unless they are exported from IS
```
GabrielKS, line 5: @jd-lara you won't love this, but exporting the aliases is the [only way](https://github.com/JuliaLang/julia/blob/master/base/show.jl#L617) way to get them to register as aliases for the purposes of printing and such, and I think overriding the printing of these types is uglier than adding them as exports. @daniel-thom agrees that this is fairly harmless — a name collision is unlikely because these would be imported rather than redefined in any package that also exports, and also we are (possibly accidentally) [already exporting](https://github.com/NREL-Sienna/InfrastructureSystems.jl/issues/391) some things from IS.

jd-lara: uggg. Fine

jd-lara (approved):

PR #393, opened 2024-08-19
Add missing or move incorrect semicolons

kdayday:
Missing or incorrect semicolon placement in function definitions was leading to multiple signatures in the documentation

May break calls for these functions:
- `open_file_logger`
- `list_recorder_events`
- `show_recorder_events`
- `get_time_series_multiple`
- `read_time_series`
- `get_time_series`
- `get_time_series_array`
- `get_time_series_timestamps`
- `get_time_series_values`


kdayday:
Tagging @daniel-thom in case this broke any tests

jd-lara (commented):
@kdayday you need to update the calls to the functions you broke as well. The tests are failing because the call to open the file logger broke with the change made in `logging.jl`

PR #396, opened 2024-09-09
Jd/fix semicolon inconsistency

jd-lara:


jd-lara:
@kdayday I will close this, after reviewing with @daniel-thom many of these corrections are actually correct use cases.

PR #399, opened 2024-09-24
Resolve `IS` prefix in submodules (fixes #398)

### LINKED ISSUE #398: `IS` is not defined within submodules
There are a few references to `IS` within `IS.Optimization`, e.g.

https://github.com/NREL-Sienna/InfrastructureSystems.jl/blob/02e36888605cb4aa6a2607b5690f9d1380379089/src/Optimization/optimization_problem_results.jl#L730

. However, `IS` is not defined in this scope. I think this hasn't been caught before because it is only in special cases that we try to reference `IS`.

GabrielKS:
Pretty straightforward: I removed the `IS` prefix in submodules where it was not necessary and added `IS` aliases in each submodule for the cases where it is necessary. Right now there are no references to `IS` in `IS.Simulation` but I think it's probably a good practice to have the alias for the future anyway. Fixes https://github.com/NREL-Sienna/InfrastructureSystems.jl/issues/398.

### DIFF
--- a/src/Optimization/Optimization.jl
+++ b/src/Optimization/Optimization.jl
@@ -10,6 +10,9 @@ import Dates
 import CSV
 import DataFrames
 
+import ..InfrastructureSystems
+const IS = InfrastructureSystems
+
 import ..InfrastructureSystems:
     @scoped_enum,
     InfrastructureSystemsType,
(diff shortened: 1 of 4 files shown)


jd-lara (approved):
Looks I left some of this code behind.

Issue #407, opened 2024-10-28
`compare_values` incorrectly reports failures on `Array`s


GabrielKS:
The following currently reports that the values are different, even though the values are the same (under `==`, `isequivalent`, everything that matters):
```julia
InfrastructureSystems.compare_values([0 0], [0 0])
```

This bug predates [my recent `compare_values` PR](https://github.com/NREL-Sienna/InfrastructureSystems.jl/pull/387); rather, the root cause is that we branch on whether a certain set of `fieldnames` is empty and as of Julia 1.11, `fieldnames(Array)` is no longer empty:
<img width="582" alt="Screenshot 2024-10-28 at 4 56 11 PM" src="https://github.com/user-attachments/assets/419c7b3e-a112-43e8-bc45-662083eb4913">

So really it's on us for relying on the private interface of an `Array`'s fields, though in the case of `compare_values` we can't really avoid relying on fields. My proposed fix is to add a special case for `AbstractArray` (or, rather, refurbish the one that already exists). Another, more flexible idea would be to always apply `match_fn` to a possibly composite object and only recurse into the fields if the match fails, but I don't have time for another `compare_values` deep dive right now.

PR #408, opened 2024-10-28
Fix bug 407

GabrielKS:
See bug https://github.com/NREL-Sienna/InfrastructureSystems.jl/issues/407 for a detailed description of the issue. This implements the proposed fix from that bug report, i.e., adds a special case for `AbstractArray`. I couldn't resist a bit more cleanup of now-redundant code as well.

### DIFF
--- a/src/utils/utils.jl
+++ b/src/utils/utils.jl
@@ -126,8 +126,6 @@ _fetch_match_fn(::Nothing) = isequivalent
 # Whether to stop recursing and apply the match_fn
 _is_compare_directly(::DataType, ::DataType) = true
 _is_compare_directly(::T, ::U) where {T, U} = true
-# As of 1.11, Arrays have fields we don't want to touch
-_is_compare_directly(::T, ::T) where {T <: AbstractArray} = true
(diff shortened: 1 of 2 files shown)


jd-lara (approved):
makes sense. Now array objects have ref and size as fields.

daniel-thom (approved):

jd-lara:
@GabrielKS is it expected that this breaks PSY?

GabrielKS:
> @GabrielKS is it expected that this breaks PSY?

No, that was not expected (though strictly speaking it is only breaking the PSY _tests_…). I'll get into it.

GabrielKS:
Okay, yesterday evening's implementation was fundamentally wrongheaded. We do need a special case for `AbstractArray` to avoid testing the fields, but we don't want it to go directly to the match function, we want it to iterate over the items and recurse — like we already did for `Vector`. Here I've adapted that existing `Vector` method so it works with arrays of arbitrary dimensionality. Should not break PowerSystems now.

jd-lara (approved):

PR #413, opened 2024-11-11
define RNG seed for all dependencies

jd-lara:
This PR adds a constant for the random seed to be used by any RNG in the platform

### DIFF
--- a/src/common.jl
+++ b/src/common.jl
@@ -41,6 +41,8 @@ const CONSTANT = Float64
 const LIM_TOL = 1e-6
 const XY_COORDS = @NamedTuple{x::Float64, y::Float64}
 
+const RNG_SEED = get(ENV, "SIENNA_RNG_SEED", 2017)
+
 # See https://github.com/JuliaLang/julia/issues/18485
 "An equality predicate that is `true` for `NaN, NaN` (unlike `==`) and for `-0.0, 0.0` (unlike `isequal`)"
 isequivalent(x, y) = isequal(x, y) || (x == y)


src/common.jl
```diff
@@ -41,6 +41,8 @@ const CONSTANT = Float64
 const LIM_TOL = 1e-6
 const XY_COORDS = @NamedTuple{x::Float64, y::Float64}
 
+const RNG_SEED = get(ENV, "SIENNA_RNG_SEED", 2017)
```
daniel-thom, line 44: Can you give an example of how setting the environment variable will be used? The value will be set the first time the code is compiled.

jd-lara:
@daniel-thom I am trying to make something like this where we just need to define this once. It will be useful here and in PRASInterface.jl 

https://github.com/NREL-Sienna/PowerSimulations.jl/commit/f7d5cafb50c5dce324e71d9a035556cac6a0ce7a

daniel-thom:
> @daniel-thom I am trying to make something like this where we just need to define this once. It will be useful here and in PRASInterface.jl
> 
> [NREL-Sienna/PowerSimulations.jl@f7d5caf](https://github.com/NREL-Sienna/PowerSimulations.jl/commit/f7d5cafb50c5dce324e71d9a035556cac6a0ce7a)

I'm still not following
1. Starting with the same seed will always produce the sequence. Are you saying that you want that?
2. How do you expect the value to be changed by the environment variable? It will get set at first compilation and then can never be changed.

Issue #382, opened 2024-06-24
Support for bulk addition of supplemental attributes


daniel-thom:
We currently only support adding supplemental attributes one at a time. This is likely to be slow if there are thousands of attributes because each addition causes one row to be added to an SQLite table. We support bulk addition of time series for the same reason and could do the same for supplemental attributes.

daniel-thom:
I made a prototype. Performance is ~3x faster with bulk additions. My benchmark adds 500k supplemental attributes to components.
- add 1 at a time: 24.378 seconds
- add 1000 at a time: 7.7 seconds

The code is straightforward and I'm not worried about that. There is a cost in user complexity for the bulk additions. The overall time involved here is not huge. Is it sufficient to justify the complexity (if users notice the docstrings)?

PR #416, opened 2024-12-06
Speed up accessors for time series and supplemental attributes

### LINKED ISSUE #417: Performance for adding multiple supplemental attributes IS

daniel-thom:
## Latency fixes
The existing code had a few issues that caused slowness when calling `has_time_series` or `get_time_series`:
- SQLite indexes were not present when de-serializing a PSB case. I'm actually still not sure why. They were correctly set after `to_json(sys)` followed by `System(file.json)`. The new code ensures that they are always set. This could have caused latency of up to 30-40 us on each query.
- Calling `SQLite.DBInterface.execute(db, "a sql query as string")` has an overhead of ~3-4 us. The delay seems to come from compiling the SQL statement. I'll note that Python does not suffer from this latency. Interpolating strings causes additional latency. This new code caches the statements. It also avoids string interpolations, but only for the `has_time_series` code path. This is perhaps over-engineered. If this still isn't fast enough for PSI, we need to re-think the entire idea and allow the application to defined what time series attributes should be cached.
- Calling `get_time_series` results in a SQL query to retrieve the time series metadata, which is stored as a JSON string, and then deserialize it into an object. That takes ~30 us. In most cases, this is harmless. Reading the data from the disk takes a lot longer than that. Some users might want to use the in-memory time series storage, and so might notice this delay. To solve that I split the metadata-as-json data into a separate table and added caching of the real metadata objects. De-serialization will only occur on the first read request.
- The existing code has a bug that causes `has_time_series(component, Deterministic)` to return false if the component has `DeterministicSingleTimeSeries`. This PR fixes the bug.
- The fact that we expect `has_time_series(component, Deterministic)` to return true when only DeterministicSingleTimeSeries is present causes extra latency. We could remove that latency by changing the requirement (make the user ask for the exact type stored) or add another column to the database table. We agreed to change this behavior in the next release.

## Bulk time series addition
This PR introduces a simplification to bulk time series addition. Previously, the user had to call `bulk_add_time_series!()` with an iterable of `TimeSeriesAssociation` instances. The code performed bulk database queries to check for duplicates and insert new rows. That was misguided. The code is now easier to write for users. They will do this instead:
```
begin_time_series_update(sys) do
    for loop over components and time series
        add_time_series!(sys, component, time_series)
    end
end
```
The new code is a ~10% faster than the old code. The user can still call the existing function `bulk_add_time_series!`. It redirects to the new code. We can deprecate that function now and remove it in the next major release.

## Supplemental attributes
- Improves query performance when searching for components attached to attributes or attributes attached to components.
- Adds a procedure for bulk addition of attributes, similar to that for time series.

Fixes #417 and #382


src/system_data.jl
```diff
@@ -567,7 +567,7 @@ function _transform_single_time_series!(
             horizon = params.horizon,
             time_series_type = DeterministicSingleTimeSeries,
             scaling_factor_multiplier = get_scaling_factor_multiplier(metadata),
-            internal = get_internal(metadata),
+            internal = InfrastructureSystemsInternal(),
```
daniel-thom, line 570: The previous behavior was incorrect, but was harmless. It caused the metadata of the transformed object to have the same UUID, but no code ever looked at that.

src/time_series_manager.jl
```diff
@@ -110,7 +110,11 @@ function add_time_series!(mgr::TimeSeriesManager, batch::Vector{TimeSeriesAssoci
         for uuid in new_ts_uuids
             serialize_time_series!(mgr.data_store, time_series_uuids[uuid])
         end
-        add_metadata!(mgr.metadata_store, owners, all_metadata)
+        if length(all_metadata) == 1
+            add_metadata!(mgr.metadata_store, owners[1], all_metadata[1])
```
daniel-thom, line 114: This is faster than calling the the array version.

src/supplemental_attribute_associations.jl
```diff
@@ -23,14 +23,17 @@ const SUPPLEMENTAL_ATTRIBUTE_TABLE_NAME = "supplemental_attributes"
 
 mutable struct SupplementalAttributeAssociations
     db::SQLite.DB
+    # If we don't cache SQL statements, there is a cost of 3-4 us on every query.
+    cached_statements::Dict{String, SQLite.Stmt}
```
daniel-thom, line 27: I made some minor changes to the supplemental attribute associations, but did not profile them. If the time series changes work out, I can pursue these in greater detail.

src/time_series_metadata_store.jl
```diff
@@ -1,8 +1,26 @@
...
+    has_features::Bool
+    feature_filter::Union{Nothing, String} = nothing
+end
+
 mutable struct TimeSeriesMetadataStore
     db::SQLite.DB
+    # Caching compiled SQL statements saves 3-4 us per query query.
+    # DBInterface.jl does something similar with @prepare.
+    # We need this to be tied to our connection.
+    cached_statements::Dict{String, SQLite.Stmt}
```
daniel-thom, line 18: This applies to selective queries that we run frequently.

src/time_series_metadata_store.jl
```diff
@@ -1,8 +1,26 @@
...
+    # This caching allows the code to skip some string interpolations.
+    # It is experimental for PowerSimulations, which calls has_metadata frequently.
+    # It may not be necessary. Savings are less than 1 us.
+    has_metadata_statements::Dict{HasMetadataQueryKey, SQLite.Stmt}
```
daniel-thom, line 22: This cached applies only to the `has_time_series` path, which is the most important to optimize for PSI. And is possibly over-engineered.

jd-lara:
@daniel-thom now there is a workaround in PowerSimulations to handle this case "The fact that we expect has_time_series(component, Deterministic) to return true when only DeterministicSingleTimeSeries is present causes extra latency." and any impact of that can be addressed.

daniel-thom:
> @daniel-thom now there is a workaround in PowerSimulations to handle this case "The fact that we expect has_time_series(component, Deterministic) to return true when only DeterministicSingleTimeSeries is present causes extra latency." and any impact of that can be addressed.

Currently, `get_time_series(Deterministic, ...)` will return a DeterministicSingleTimeSeries. `has_time_series` doesn't, and that is a bug. We need to make the behaviors consistent. Changing `get_time_series` could be a big API breakage for users. Let's test the performance on a large system before we make a decision.

jd-lara:
@daniel-thom I tested this branch and it should be merged. These are the results on the EI 

main: 

Build of The EI: 407.636684 seconds (3.26 G allocations: 109.041 GiB, 43.34% gc time, 17.21% compilation time: 2% of which was recompilation)

Loop 10 times over all StaticInjections and call `has_time_series` on the EI: 2939.589924 seconds (22.27 M allocations: 1.222 GiB, 0.01% gc time, 0.00% compilation time)

This branch 

Build of The EI: 384.807498 seconds (3.25 G allocations: 108.283 GiB, 44.03% gc time, 17.02% compilation time: 2% of which was recompilation)

Loop 10 times over all StaticInjections and call `has_time_series` on the EI: 1.538067 seconds (14.96 M allocations: 727.982 MiB, 5.98% gc time, 2.98% compilation time)

src/time_series_metadata_store.jl
```diff
@@ -541,85 +753,159 @@ function get_time_series_summary_table(store::TimeSeriesMetadataStore)
...
+
+    where_clauses = String[]
+    if key.has_name
+        push!(where_clauses, "name = ?")
+    end
+    if key.num_possible_types == 1
+        push!(where_clauses, "time_series_type = ?")
+    elseif key.num_possible_types > 1
+        val = chop(repeat("?,", key.num_possible_types))
+        push!(where_clauses, "time_series_type IN ($val)")
```
daniel-thom, line 837: Why isn't this tested?

jd-lara (approved):

PR #424, opened 2025-01-11
Ensure that a component cannot be added to two systems

daniel-thom:
We used to have a check like this. I discovered today that we currently don't and am restoring it. It's possible that this change will cause a problem for user scripts, but this is definitely a bug. If the user is adding a component to a second system, the first system is now inconsistent.


jd-lara (approved):

jd-lara:
@daniel-thom I added a release of PSY with the bug fix. When its merged into the registry we can re-test this PR to merge and release

PR #426, opened 2025-01-16
add serialization/deserialization support for vector of complex

rodrigomha:
Do I need to add an specific test for this?

This is required for SwitchedAdmittance struct in this PR:
https://github.com/NREL-Sienna/PowerSystems.jl/pull/1252


jd-lara (approved):

daniel-thom:
> Do I need to add an specific test for this?
> 
> This is required for SwitchedAdmittance struct in this PR: [NREL-Sienna/PowerSystems.jl#1252](https://github.com/NREL-Sienna/PowerSystems.jl/pull/1252)

It would be good to add a test. It should be very simple, something like https://github.com/NREL-Sienna/InfrastructureSystems.jl/blob/main/test/test_serialization.jl#L51

daniel-thom (approved):
LGTM

PR #429, opened 2025-01-21
Fix deserialization of supplemental attributes

daniel-thom:
The shared system references object was not getting assigned to supplemental attributes during deserialization. This meant that you could not find time series attached to attributes afterwards. This has likely been broken for many months.


jd-lara (approved):
LGTM

PR #433, opened 2025-01-28
Disallow normalization factors of 0

### LINKED ISSUE #414: Time series Normalization Factor = 0
When using "normalization_factor": "Max", it takes the maximum value of the time series as a normalization factor. 

However, if all the values of a time series are 0, then it results in a time series loaded as "NaN". In such cases, this is supposed to be detected and handled to avoid time series filled wit  "NaN".

Could be possible to add an option to normalize taking the _active_power_limits__?

daniel-thom:
Fixes #414


jd-lara:
This situation will still tigger errors but I think it makes more sense and it is more informative to the user.

PR #434, opened 2025-01-31
Fix the type of horizon_ms in the time series metadata database

daniel-thom:
The code was mistakenly writing `horizon_ms` as a serialized `Dates.Millisecond` object instead of an integer. This PR fixes the database during deserialization.

Bug introduction: when we converted horizon from an integer to a period on 2024-05-13.

Impact of the bug: usually, none. However, there was one case this week where a user had a custom environment and adding  a row to the db failed because of what was likely a different in the version of the Dates package.

### DIFF
--- a/src/time_series_metadata_store.jl
+++ b/src/time_series_metadata_store.jl
@@ -59,6 +59,7 @@ function TimeSeriesMetadataStore(filename::AbstractString)
         _migrate_legacy_schema(store)
     end
 
+    _apply_horizon_type_fix_if_needed(store.db)
     _load_metadata_into_memory!(store)
     _create_indexes!(store)
     @debug "Loaded time series metadata from file" _group = LOG_GROUP_TIME_SERIES filename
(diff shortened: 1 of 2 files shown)


src/time_series_metadata_store.jl
```diff
@@ -78,6 +79,37 @@ function _needs_migration(db::SQLite.DB)
...
+            FROM $ASSOCIATIONS_TABLE_NAME
+            WHERE horizon_ms IS NOT NULL
+        """
+        update_query = "UPDATE $ASSOCIATIONS_TABLE_NAME SET horizon_ms = ? WHERE id = ?"
+        stmt = SQLite.Stmt(db, update_query)
+        SQLite.transaction(db) do
+            for row in Tables.rowtable(SQLite.DBInterface.execute(db, query))
+                if row.horizon_ms isa Dates.Period
+                    params = (row.horizon_ms.value, row.id)
+                    SQLite.DBInterface.execute(stmt, params)
```
josephmckinsey, line 106: Should we be using executemany here?

daniel-thom: DBInterface.jl says this for executemany:
```
By default, each set of parameters will be looped over and DBInterface.execute will be called for each. 
```
https://juliadatabases.org/DBInterface.jl/dev/#DBInterface.executemany

So, I don't see any value in building an intermediate vector. This seems most natural. I'm open to a counter opinion.

src/time_series_metadata_store.jl
```diff
@@ -1140,7 +1172,7 @@ function _create_row(
         ts_category,
         string(get_initial_timestamp(metadata)),
         Dates.Millisecond(get_resolution(metadata)).value,
-        Dates.Millisecond(get_horizon(metadata)),
+        Dates.Millisecond(get_horizon(metadata)).value,
```
daniel-thom, line 1175: I should say that @josephmckinsey found this bug after debugging an error from @hannahchubin. I was surprised to learn
1. When SQLite.jl tries to write a value to SQLite, it will be direct pass thru if there is an obvious mapping (integer to integer). If a mapping doesn't exist, it will call `Serialization.serialize(value)` and store that as a BLOB in SQLite. The documentation does not make this obvious. The comments in the code make it clear that this may not work across Julia versions. I would have expected an error and be forced to opt-in to this behavior.
2. SQLite will happily store the BLOB in a column that you've defined as an integer.

Am I questioning past decisions? Maybe. I'm certainly not happy about this. Thanks @josephmckinsey.

josephmckinsey (approved):

jd-lara (approved):
This PR looks fine to me. But the conflicts need to be resolved

PR #435, opened 2025-01-31
Drop the metadata table when it's not needed

daniel-thom:
This PR changes the behavior of the time series metadata store regarding the time series metadata struct values. 

Existing behavior: Keep a table with columns `metadata_uuid` and `metadata` (which is the JSON serialized value). Lazily load the metadata objects into memory whenever the user reads a time series.

New behavior: Store that metadata table in the SQLite database at serialization time. During deserialization, load the contents of that table into memory and drop the table.

Reason for change: Lower memory usage and faster additions of new time series because we only add to one table instead of two.

### DIFF
--- a/src/time_series_metadata_store.jl
+++ b/src/time_series_metadata_store.jl
@@ -37,7 +37,6 @@ function TimeSeriesMetadataStore()
         Dict{Base.UUID, TimeSeriesMetadata}(),
     )
     _create_associations_table!(store)
-    _create_metadata_table!(store)
     _create_indexes!(store)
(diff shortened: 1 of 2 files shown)


jd-lara (approved):

jd-lara:
The changes make sense. @daniel-thom should we benchmark with EI to get metrics?

daniel-thom:
> The changes make sense. @daniel-thom should we benchmark with EI to get metrics?

No need. It will be a tiny bit faster. No one will likely notice. There is no risk of being slower at runtime. Serde might be a tad bit slower, but I'm sure people will take the tradeoff.

PR #436, opened 2025-02-03
Fix supplemental attribute table when component UUID is changed

daniel-thom:


jd-lara (approved):
This is a sensible change. I don't think we should be encouraging users to make UUID changes anyway.

PR #438, opened 2025-02-11
Update OptimizationResults to use `Vector{DateTime}`

rodrigomha:
New version for #415 as rebasing was a mess.

- Added a test that check that using the previous StepRange works fine, as it does the conversion automatically.
- Also added code that still output the result if the length is different than timesteps. Give a warning when that happens. The test also check that timestamps are ignored in that case.

### DIFF
--- a/src/Optimization/optimization_problem_results.jl
+++ b/src/Optimization/optimization_problem_results.jl
@@ -17,20 +17,20 @@ mutable struct OptimizationProblemResults <: Results
 end
 
 function OptimizationProblemResults(
-    base_power::Float64,
+    base_power,
     timestamps::StepRange{Dates.DateTime, Dates.Millisecond},
-    source_data::Union{Nothing, InfrastructureSystemsType},
-    source_data_uuid::Base.UUID,
-    aux_variable_values::Dict{AuxVarKey, DataFrames.DataFrame},
-    variable_values::Dict{VariableKey, DataFrames.DataFrame},
-    dual_values::Dict{ConstraintKey, DataFrames.DataFrame},
-    parameter_values::Dict{ParameterKey, DataFrames.DataFrame},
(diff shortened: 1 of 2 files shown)


rodrigomha:
This is needed for the new SiennaInvest PR

src/Optimization/optimization_problem_results.jl
```diff
@@ -1,7 +1,7 @@
 # This needs renaming to avoid collision with the DecisionModelResults/EmulationModelResults
 mutable struct OptimizationProblemResults <: Results
```
jd-lara, line 2: we still need a method that takes the range to avoid breaking PSI.

src/Optimization/optimization_problem_results.jl
```diff
@@ -58,6 +91,16 @@ function get_objective_value(res::OptimizationProblemResults, execution = 1)
     return res.optimizer_stats[execution, :objective_value]
 end
 
+function get_resolution(res::OptimizationProblemResults)
+    # Method return the first resolution between timestamps
+    # If single timestamp is used, it return nothing
+    diff_res = diff(get_timestamps(res))
+    if !isempty(diff_res)
+        return first(diff_res)
```
GabrielKS, line 99: Are we expecting that all the time increments should be the same? If so:
```suggestion
        return only(unique(diff_res))
```

If not, we should document clearly somewhere that `get_resolution` only gets the first increment, that seems like a surprising behavior

rodrigomha: I added the warning for the case of multiple resolutions.

We want to support being able to have multiple resolutions, but keep the structure mostly the same for PowerSimulations.

jd-lara:
we need an strategy for this feature that it isn't breaking in PSI... it concerns me that it could break if the user improperly updates IS and not PSI creating a cery hard to track error.

PR #442, opened 2025-02-21
Add resolution to time series methods

daniel-thom:
Needed by https://github.com/NREL-Sienna/PowerSimulations.jl/issues/1170


src/time_series_metadata_store.jl
```diff
@@ -892,52 +903,6 @@ function get_time_series_resolutions(
     )
 end
 
-"""
```
daniel-thom, line 895: This was not being called.

jd-lara (approved):

PR #444, opened 2025-03-05
Fix bug for read results method with kwargs len or start_time

rodrigomha:
The current approach was checking the length of the preprocessed time stamps (if the users want to specify the length or start time, or both) to ignore printing the timestamps for variables with different length. The correct approach should check the base timestamps (and not the preprocessed) to do the read result.

Closes https://github.com/NREL-Sienna/PowerSimulations.jl/issues/1267

### DIFF
--- a/src/Optimization/optimization_problem_results.jl
+++ b/src/Optimization/optimization_problem_results.jl
@@ -310,6 +310,7 @@ function _read_results(
     timestamps::Vector{Dates.DateTime},
     time_ids,
     base_power::Float64,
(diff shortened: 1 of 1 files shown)


GabrielKS (approved):
Can confirm that this resolves https://github.com/NREL-Sienna/PowerSimulations.jl/issues/1267 and the related PowerAnalytics test failures. It would be great if we could get some unit tests somewhere in either IS or PSI that test this behavior.

PR #448, opened 2025-03-11
fix: Cascading removing of component_association from remove_component

### LINKED ISSUE #447: Bug: removing a component that has a supplemental attribute is not removing the association.
We looked at the code and when we call `remove_component` we only check that if it has time series, but we do not remove any existing associations to supplemental attributes.

pesap:
Closes #447

### DIFF
--- a/test/test_supplemental_attributes.jl
+++ b/test/test_supplemental_attributes.jl
@@ -116,14 +116,23 @@ end
 
 @testset "Test remove supplemental_attribute association on remove_component" begin
     data = IS.SystemData(; time_series_in_memory = true)
-    geo_supplemental_attribute = IS.GeographicInfo()
+    mgr = data.supplemental_attribute_manager
+    attribute = IS.GeographicInfo()
     component1 = IS.TestComponent("component1", 5)
+    component2 = IS.TestComponent("component2", 6)
     IS.add_component!(data, component1)
-    IS.add_supplemental_attribute!(data, component1, geo_supplemental_attribute)
+    IS.add_component!(data, component2)
+    IS.add_supplemental_attribute!(data, component1, attribute)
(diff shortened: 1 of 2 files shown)


daniel-thom (approved):
LGTM. Thanks.

test/test_supplemental_attributes.jl
```diff
@@ -114,6 +114,18 @@ end
...
 
+@testset "Test remove supplemental_attribute association on remove_component" begin
+    data = IS.SystemData(; time_series_in_memory = true)
+    geo_supplemental_attribute = IS.GeographicInfo()
+    component1 = IS.TestComponent("component1", 5)
+    IS.add_component!(data, component1)
+    IS.add_supplemental_attribute!(data, component1, geo_supplemental_attribute)
+    @test IS.get_num_supplemental_attributes(data) == 1
+
+    IS.remove_component!(component1)
```
daniel-thom, line 125: missing the system

pesap: I  made the test better.

GabrielKS (commented):
`remove_component!` has the kwarg `remove_time_series` — looks like the use case for that is `mask_component!`, which removes a component from the main components collection and adds it instead to a masked components collection with time series intact. Do we want a `remove_supplemental_attributes` kwarg that does the same thing?

pesap:
> `remove_component!` has the kwarg `remove_time_series` — looks like the use case for that is `mask_component!`, which removes a component from the main components collection and adds it instead to a masked components collection with time series intact. Do we want a `remove_supplemental_attributes` kwarg so we can keep supplemental attributes too in this case?

I think that makes sense to me. Shall we open a new issue for that and make the change on a separate PR?

PR #453, opened 2025-03-21
Preserve supplemental attributes in mask_component!

### LINKED ISSUE #449: Add new kwarg to `remove_component` that removes supplemental attribute
Proposal
---------

```diff
function _remove_component!(
    ::Type{T},
    components::Components,
    name::AbstractString;
    remove_time_series = true,
+ remove_supplemental_attribute = true,
) where {T <: InfrastructureSystemsComponent}
```
Context
--------              
`remove_component!` has the kwarg `remove_time_series` — looks like the use case for that is `mask_component!`, which removes a component from the main components collection and adds it instead to a masked components collection with time series intact. Do we want a `remove_supplemental_attributes` kwarg so we can keep supplemental attributes too in this case?

_Originally posted by @GabrielKS in https://github.com/NREL-Sienna/InfrastructureSystems.jl/pull/448#pullrequestreview-2676362910_

daniel-thom:
Fixes #449

### DIFF
--- a/src/components.jl
+++ b/src/components.jl
@@ -138,12 +138,14 @@ function remove_component!(
     components::Components,
     component::T;
     remove_time_series = true,
+    remove_supplemental_attributes = true,
 ) where {T <: InfrastructureSystemsComponent}
     return _remove_component!(
         T,
         components,
         get_name(component);
         remove_time_series = remove_time_series,
+        remove_supplemental_attributes = remove_supplemental_attributes,
     )
 end
 
@@ -157,8 +159,15 @@ function remove_component!(
     components::Components,
     name::AbstractString;
     remove_time_series = true,
+    remove_supplemental_attributes = true,
 ) where {T <: InfrastructureSystemsComponent}
-    return _remove_component!(T, components, name; remove_time_series = remove_time_series)
+    return _remove_component!(
+        T,
+        components,
+        name;
+        remove_time_series = remove_time_series,
+        remove_supplemental_attributes = remove_supplemental_attributes,
+    )
 end
 
 function _remove_component!(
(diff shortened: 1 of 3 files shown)


pesap (approved):
LGTM. Thanks, @daniel-thom !

GabrielKS (commented):
Implementation looks good. But @daniel-thom [noted](https://github.com/NREL-Sienna/InfrastructureSystems.jl/issues/449#issuecomment-2744252991) that "you cannot have a supplemental attribute in the system that is not connected to a component"; this would seem to allow a user to break that rule. Do we want to add a layer of indirection such that `mask_component!` can set `remove_supplemental_attributes = false` but the user can't?

daniel-thom:
> Implementation looks good. But @daniel-thom [noted](https://github.com/NREL-Sienna/InfrastructureSystems.jl/issues/449#issuecomment-2744252991) that "you cannot have a supplemental attribute in the system that is not connected to a component"; this would seem to allow a user to break that rule. Do we want to add a layer of indirection such that `mask_component!` can set `remove_supplemental_attributes = false` but the user can't?

Is there a case where the user has access to this behavior? We don't provide this parameter in the `PSY.System` version of `remove_component`.

GabrielKS:
> Is there a case where the user has access to this behavior? We don't provide this parameter in the `PSY.System` version of `remove_component`.

Oh, I guess I was just thinking of `IS.remove_component!` as public interface of IS.

daniel-thom:
> > Is there a case where the user has access to this behavior? We don't provide this parameter in the `PSY.System` version of `remove_component`.
> 
> Oh, I guess I was just thinking of `IS.remove_component!` as public interface of IS.

I don't think we need a new abstraction. There are likely other cases similar to this. IS.SystemData provides developer access, not user access.

GabrielKS (approved):

PR #455, opened 2025-03-24
Support irregular time series resolutions

daniel-thom:
This draft PR demonstrates how we could support irregular time series resolutions, such as `Dates.Month` and `Dates.Year`. The main problem with the current code is that it converts timestamps in a time series array to indexes by calculating `(t2 - t1) / resolution`, which doesn't work with subtypes of `Dates.Period` that cannot be converted to milliseconds.

There are a few questions that need resolution before we can proceed.
1. When a user passes a `TimeSeries.TimeArray` to the `SingleTimeSeries` constructor, we compute the resolution for the user by subtracting the first two timestamps. Can we require that the user pass resolution when that resolution is irregular? Otherwise, we would have to make a guess when the resolution is about a month or year, or some multiple of those.
2. Can the fields `horizon` and `interval` also be irregular?

## TODOs
- [ ] Perform database migration because this code changes the data type for resolution, and maybe horizon and interval.
- [ ] Add tests for all cases 
- [ ] TODOs in the code

### DIFF
--- a/src/deterministic.jl
+++ b/src/deterministic.jl
@@ -65,6 +65,9 @@ Construct Deterministic from a Dict of TimeArrays.
 
   - `name::AbstractString`: user-defined name
   - `input_data::AbstractDict{Dates.DateTime, TimeSeries.TimeArray}`: time series data.
+  - `resolution::Union{Nothing, Dates.Period} = nothing`: If nothing, infer resolution from the
+    data. Otherwise, this must be the difference between each consecutive timestamps. This is
+    required if the resolution is not constant, such as Dates.Month or Dates.Year.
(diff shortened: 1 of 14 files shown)


jd-lara (commented):
@daniel-thom what's the status of this PR?

rodrigomha:
Using `Printf = 1.11.0` in the toml forces the user to have Julia 1.11. Do we need to force Julia 1.11 for this PR? Or we can live with `Printf = 1` ?

jd-lara (approved):
LGTM. I didn't see anything that caught my attention as problematic. More testing before release will help

PR #457, opened 2025-03-28
add function to get random seed

jd-lara:
This is a step to make sure that over the whole platform the same random seed is used and allows the user to create an ENV variable for reproducibility if they want to.

### DIFF
--- a/src/random_seed.jl
+++ b/src/random_seed.jl
@@ -1,5 +1,3 @@
-const SIENNA_RANDOM_SEED = 07012017
-
 function get_random_seed()
     if haskey(ENV, "SIENNA_RANDOM_SEED")
         try
@@ -9,6 +7,7 @@ function get_random_seed()
             @error("SIENNA_RANDOM_SEED $val can't be read as an integer value")
             rethrow(e)
         end
+        return SIENNA_RANDOM_SEED
     end
-    return SIENNA_RANDOM_SEED
+    return rand(Int)
 end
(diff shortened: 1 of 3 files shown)


GabrielKS:
If we go with this approach, might be convenient to have a setter too.

jd-lara:
> If we go with this approach, might be convenient to have a setter too.

I don't think that we can set a env variable from a function like this

src/random_seed.jl
```diff
@@ -0,0 +1,13 @@
+function get_random_seed()
+    if haskey(ENV, "SIENNA_RANDOM_SEED")
+        try
+            return parse(Int, ENV["SIENNA_RANDOM_SEED"])
+        catch e
+            val = ENV["SIENNA_RANDOM_SEED"]
+            @error("SIENNA_RANDOM_SEED $val can't be read as an integer value")
+            rethrow(e)
+        end
+        return SIENNA_RANDOM_SEED
```
daniel-thom, line 10: Should this be in an else clause?

src/random_seed.jl
```diff
@@ -0,0 +1,13 @@
+function get_random_seed()
+    if haskey(ENV, "SIENNA_RANDOM_SEED")
+        try
+            return parse(Int, ENV["SIENNA_RANDOM_SEED"])
+        catch e
+            val = ENV["SIENNA_RANDOM_SEED"]
+            @error("SIENNA_RANDOM_SEED $val can't be read as an integer value")
+            rethrow(e)
```
daniel-thom, line 8:
```suggestion
            rethrow()
```

src/random_seed.jl
jd-lara: I think it is equivalent but I can put it in an else clause

daniel-thom (approved):

PR #458, opened 2025-03-28
Implement non-allocating `@scoped_enum`

GabrielKS:
The fundamental problem behind https://github.com/NREL-Sienna/PowerSystems.jl/pull/1216 is that `@scoped_enum` is backed by a `Dict`, which is mutable, so we can't compile away the lookup even in cases where it looks like a constant (e.g., `my_units == UnitSystem.NATURAL_UNITS`). There is no notion of a "constant dictionary" in Julia. However, a dictionary is just a function with a very small domain, and `Val`-based function application is more amenable to compile-time optimization. Here, I use that idea to implement a non-allocating `@scoped_enum`. This should eliminate the need for `@scoped_enum` users to implement their own patchwork fixes like https://github.com/NREL-Sienna/PowerSystems.jl/pull/1216 when they need performance.

Draft because I'm not confident that this is the right approach: I've tested that it doesn't allocate, but I still need to test other aspects of performance. Also, this is not how `@enum` is implemented in `@Base`, maybe an implementation along those lines would be even better.

### DIFF
--- a/src/utils/utils.jl
+++ b/src/utils/utils.jl
@@ -256,6 +256,15 @@ julia> @scoped_enum(Fruit,
 ```
 """
 macro scoped_enum(T, args...)
+    hn_methods = []
+    n2v_methods = []
+    v2n_methods = []
+    for p in args
+        value = Int64(last(p.args))
+        push!(hn_methods, :(_hasname(::$(Val{first(p.args)})) = true))
+        push!(n2v_methods, :(_name2value(::$(Val{first(p.args)})) = $value))
+        push!(v2n_methods, :(_value2name(::Val{$value}) = $(String(first(p.args)))))
+    end
     blk = esc(
         :(
             module $(Symbol("$(T)Module"))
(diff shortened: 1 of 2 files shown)


jd-lara:
@daniel-thom do you remember why we went with @scoped_enum? I don't recall why we went that way

GabrielKS:
Okay, this is ready for review. It doesn't solve _all_ the issues with, for instance, rapidly getting a bunch of values that need to be unit converted from PowerSystems — there's some type instability there that will need to be fixed elsewhere — but as far as I can tell it will not impede such a fix and it _does_ fix anything with `ACBusTypes` we might be doing in PowerFlows. If you run the included performance tests with the old version of `@scoped_enum`, you can see how much allocation this saves in certain cases.

josephmckinsey:
So `EnumX.jl` will kind of work, as long as we accept bitcasting internally, and we figure out the de/serialization

```julia
@enumx Fruit APPLE = 1 ORANGE = 2

@testset "Test scoped_enum performance" begin
    is_apple(x::Fruit.T) = x == Fruit.APPLE
    function f()
        @test (@allocated Fruit.APPLE) == 0
        @test (@allocated instances(Fruit.T)) == 0

        rng = Random.Xoshiro(47)
        my_fruits = [rand(rng, instances(Fruit.T)) for _ in 1:1_000]
        my_results = Vector{Bool}(undef, length(my_fruits))
        # Ref(x) rather than [x] is necessary to avoid spurious allocation
        @test (@allocated my_results .= (my_fruits .== Ref(Fruit.APPLE))) == 0
        # After compilation, here is observed the most drastic difference between the Dict-based implementation and the multiple dispach-based one:
        @test (@allocated my_results .= is_apple.(my_fruits)) == 0
        @test (@allocated my_results .= (my_fruits .< Ref(Fruit.ORANGE))) == 0
        @test (@allocated my_fruits .= Fruit.T.(Int32(3) .- Base.bitcast.(Int32, my_fruits))) == 0
        @test (@allocated(
            my_fruits .= Base.bitcast.(Fruit.T, Int32(3) .- Base.bitcast.(Int32, my_fruits)))) == 0
    end
    f()
    f()
end
```

jd-lara (approved):

GabrielKS:
Let's merge this for now. It doesn't commit us to anything — we can easily switch to something else if we find something better.

PR #462, opened 2025-05-05
add fuel offtake

### LINKED ISSUE #461: ability to separate start costs into fuel-related and non-fuel related costs
I would like the ability to define startup related costs using two attributes. startup costs that are not fuel related and startup costs that are fuel-related. for the fuel-related start costs it would be great if user can define a MMBTU amount by start type (i.e., hot, warm, cold) and PSY/PSI would then apply that fuel amount times the current fuel price active for that time step to come up with a dollar amount that would then be added to the non-fuel related start costs to arrive at an all-in start costs

jd-lara:
closes #461

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,9 @@
+repos:
+  - repo: local
+    hooks:
+      - id: julia-formatter
+        name: Run Julia formatter
+        entry: julia scripts/formatter/formatter_code.jl
+        language: system
+        types: [file]
+        pass_filenames: false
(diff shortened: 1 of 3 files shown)


.pre-commit-config.yaml
```diff
@@ -0,0 +1,9 @@
+repos:
```
daniel-thom, line 1: Is that an overall Sienna workflow change (I am in favor). Do we need developer instructions? We could also solve this in a way that does not require a Python installation.

jd-lara: We are rolling it out slowly and see how it goes. We can add this to the instructions, for me using home-brew was a very easy thing to do.

pesap (approved):
Looks good to me. I added the follow-up issue on the python side. https://github.com/NREL/infrasys/issues/79?issue=NREL%7Cinfrasys%7C89

Issue #446, opened 2025-03-11
Proposal: Add `convert` methods for mixed-ownership `Union`s to reduce constructor duplication


GabrielKS:
Before https://github.com/NREL-Sienna/PowerSystems.jl/pull/1297, `PSY.MarketBidCost.shut_down` was of type `Float64`. If you called the constructor with an int, the [`convert` machinery](https://docs.julialang.org/en/v1/manual/conversion-and-promotion/) would automatically convert it to a float. That PR makes `shut_down` a `Union{TimeSeriesKey, Float64}` and that automatic process breaks down:

```julia
ERROR: MethodError: Cannot `convert` an object of type 
  Int64 to an object of type 
  Union{Float64, TimeSeriesKey}
The function `convert` exists, but no method is defined for this combination of argument types.
```

because it doesn't automatically know whether it should convert to `Float64` or `TimeSeriesKey`. See https://discourse.julialang.org/t/union-type-in-struct-type-conversion/42276 and https://github.com/JuliaLang/julia/issues/51697.

Typically when this happens, we add a bunch of auxiliary constructors to manually perform the conversions, and that's what I did in that PR. But this process is cumbersome and error-prone, is not compatible with `@kwdef`, and it takes some care to avoid method ambiguities in the case where you might be manually performing multiple conversions for different fields of the same instance.

I claim that for union types where we _own at least one of the constituent types and depend on the rest_, it is not type piracy to define things like
```julia
Base.convert(::Type{Union{Float64, TimeSeriesKey}}, x::Integer) = convert(Float64, x)

Base.convert(::Type{Union{Nothing, Float64, TimeSeriesKey}}, x::Integer) =
    Base.convert(::Type{Union{Nothing, Float64}, x})  # RHS method is already defined in Base
```
: this is just saying "it doesn't convert to one of our types, it must convert to one of yours." I propose that we define such methods for common Sienna union types, in the packages where the constituent types we own are defined, to eliminate the need for so many of these auxiliary constructors.

jd-lara:
closing as irrelevant in IS4 with TimeVariableFunctions re-design

PR #465, opened 2025-05-15
fix the deserialization for some fields in HydroReservoir

rbolgaryn:


src/serialization.jl
```diff
@@ -252,6 +252,13 @@ end
 
 deserialize(::Type{Dates.DateTime}, val::AbstractString) = Dates.DateTime(val)
 
+# TimeSeriesKey is covered by deserialize(T::Union, data::Dict) above
+deserialize(::Type{Union{Float64, TimeSeriesKey}}, val::Integer) = Float64(val)
```
daniel-thom, line 256: There is likely a better solution. Can we discuss it?

pesap (approved):
LGTM. Will let @daniel-thom  to take the lead on when to close this.

src/serialization.jl
daniel-thom: I think I prefer the solutions discussed by @GabrielKS in https://github.com/NREL-Sienna/InfrastructureSystems.jl/issues/446. He added constructors to handle the Float64/Integer conversion. The HydroReservoir struct has the issue with all of these fields:
```
    outflow::Union{Float64, TimeSeriesKey}
    level_targets::Union{Nothing, Float64, TimeSeriesKey}
    head_to_volume_factor::Union{Float64, PiecewisePointCurve}
```
The `convert` proposal would solve all of them.

I would like to ask a broader question: Is there a better way to store this information? These type of unions force a lot of complexity on applications.
- Every call site has to handle the different types.
- Serialization/de-serialization is much more complex. Auto-generated OpenAPI clients, in particular, end up with a lot of non-trivial logic to handle it.

Does SiennaGridDB already handle this situation? @josephmckinsey

jd-lara: I think @GabrielKS has a proposal to move away from the unions for PSY 6. Still TBD.

jd-lara (commented):
The fact that these fields are time varying should be handled in HPS not in the structure. Upon a second look we need to not have these be unions.

jd-lara:
checking with Pedro this will be fixed in https://github.com/NREL-Sienna/PowerSystems.jl/pull/1394

src/serialization.jl
josephmckinsey: OpenAPI.jl can deserialize a oneOf of two different types as long as discriminator is used somewhere. Right now, we haven't tackled TimeSeriesKey, but we have done PiecewisePointCurve and Float64, so we handle the head_to_volume_factor.

GabrielKS:
@jd-lara @daniel-thom I think #446 ought to be implemented no matter what, but I don't think it absolutely has to happen now; hopefully should be able to be done backwards compatibly. Separately, yes I think we should redesign the way we handle things that are optionally time series (currently `Union{TimeSeriesKey, something_else}`), but that's likely a much larger project and I don't yet have a fully thought out alternative; will definitely have to wait for PSY6.

PR #468, opened 2025-05-21
Add list_associated_component_uuids with filter by component type

### LINKED ISSUE #467: Add method get_associated_components with a component type filter
```
get_associated_components(sys, supplemental_attribute, component_type)
```

daniel-thom:
Fixes #467

### DIFF
--- a/src/system_data.jl
+++ b/src/system_data.jl
@@ -1089,6 +1089,37 @@ function get_associated_components(
     ]
 end
 
+"""
+Return all components associated with the attribute that match `component_type`.
+
+# Arguments
+- `component_type::{Type::<:InfrastructureSystemsComponent}`: Required type of the
+components to return. Can be concrete or abstract.
+- `data::SystemData`: the `SystemData` to search
+- `attribute::SupplementalAttribute`: Only return components associated with this attribute.
+- `only_available::Bool`: If true, only return components that are available.
+"""
+function get_associated_components(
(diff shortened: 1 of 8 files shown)


src/system_data.jl
```diff
@@ -1089,6 +1089,59 @@ function get_associated_components(
     ]
 end
 
+"""
```
daniel-thom, line 1092: I think we should be using a function like the one in this comment. I did not add it because of the outcome of a conversation I had with @GabrielKS: it would be confusing to users who currently use get_components/get_available_components.

```julia
function get_associated_components(
    component_type::Type{<:InfrastructureSystemsComponent},
    data::SystemData,
    attribute::SupplementalAttribute;
    only_available::Bool = false,
)
    components = component_type[]
    for uuid in list_associated_component_uuids(
        data.supplemental_attribute_manager.associations,
        attribute,
        component_type,
    )
        component = get_component(data, uuid)
        if !only_available || get_available(component)
            push!(components, component)
        end
    end

    return components
end
```
Tracked in #469

test/events.jl
```diff
@@ -0,0 +1,19 @@
+struct TestEvent <: IS.AbstractRecorderEvent
```
daniel-thom, line 1: Forgive this unrelated change. It was easy to fix alongside the addition of `test/components.jl`.

GabrielKS (approved):
LGTM!

PR #471, opened 2025-05-28
Add get_component_supplemental_attribute_pairs

daniel-thom:


src/system_data.jl
```diff
@@ -1142,6 +1142,34 @@ function get_available_associated_components(
     return components
 end
 
+"""
+Return a vector of NamedTuples with pairs of components and supplemental attributes that
+are associated with each other.
+"""
+function get_component_supplemental_attribute_pairs(
```
GabrielKS, line 1149: It's not _quite_ as evident to me that we need a `get_available...` here as in your previous case https://github.com/NREL-Sienna/InfrastructureSystems.jl/pull/468/files#r2103334665, but seems like we should still probably have one? I could be convinced otherwise.

test/test_system_data.jl
```diff
@@ -788,6 +788,56 @@ end
...
+        PVGenerator,
+        IS.SupplementalAttribute,
+        data;
+        only_available_components = true,
+    )
+    @test length(ca_pairs) == 0
+
+    ca_pairs = IS.get_component_supplemental_attribute_pairs(
+        ThermalGenerator,
+        IS.SupplementalAttribute,
```
GabrielKS, line 833: nitpick: test with concrete supplemental attribute type

src/system_data.jl
```diff
@@ -1142,6 +1142,34 @@ function get_available_associated_components(
     return components
 end
 
+"""
+Return a vector of NamedTuples with pairs of components and supplemental attributes that
+are associated with each other.
```
GabrielKS, line 1147: Is it worth documenting the names in the `NamedTuple`?

GabrielKS (changes requested):

test/test_system_data.jl
daniel-thom: Oops, I thought I had one of those. Thanks.

daniel-thom:
> Dann, this looks great to me!.

@SebastianManriqueM Do you want to keep the functionality I added in #468? That function implies that you will (1) identify supplemental attributes, (2) iterate over them, (3) on each iteration make a call to get associated components.

This new method should be significantly faster because it makes only one call to the backend database. The workflow may also be preferable, I'll defer to you. You would request up front the component and attribute types you want. If this function needs to be extended to allow custom filters on either components or attributes, we could do that.

If the functionality is redundant, I would vote for removing #468. We haven't yet made a release, and so it is safe to remove.

src/component_selector.jl
```diff
@@ -146,7 +146,7 @@ Use `optional_and_fns` to return a function that computes the conjunction of `ge
 and the given function
 """
 available_and_fn(fn::Union{Function, Nothing}, sys) =
-    optional_and_fns(Base.Fix1(get_available, sys), fn)
```
daniel-thom, line 149: @GabrielKS You'll want to review these changes. Feel free to disagree.

src/system_data.jl
```diff
@@ -1142,6 +1142,47 @@ function get_available_associated_components(
     return components
 end
 
+"""
```
daniel-thom, line 1145: @jd-lara @GabrielKS I now consider the interface finalized. I decided against adding a filter function. We could add one with the signature `function x(::InfrastructureSystemsComponent, ::SupplementalAttribute)::Bool`. Seems like overkill to me, but I'm willing to add it if you think people will want it.

src/component_container.jl
```diff
@@ -32,25 +31,22 @@ function get_components end
...
-
 "Like `get_components` but only on components that are available."
 function get_available_components end
 
 "Like `get_component` but only on components that are available."
 function get_available_component end
 
 get_available_components(sys::ComponentContainer, args...; kwargs...) =
-    get_components(c -> get_available(sys, c), sys, args...; kwargs...)
+    get_components(c -> get_available(c), sys, args...; kwargs...)
```
jd-lara, line 41: get_available always returns true/false. There is no need to create a new anonymous function 
```suggestion
    get_components(get_available, sys, args...; kwargs...)
```

src/component_container.jl
```diff
@@ -32,25 +31,22 @@ function get_components end
...
 
 get_available_components(
     ::Type{T},
     sys::ComponentContainer,
     args...;
     kwargs...,
 ) where {T <: InfrastructureSystemsComponent} =
-    get_components(c -> get_available(sys, c), T, sys, args...; kwargs...)
+    get_components(c -> get_available(c), T, sys, args...; kwargs...)
```
jd-lara, line 49:
```suggestion
    get_components(get_available, T, sys, args...; kwargs...)
```

src/component_selector.jl
jd-lara: I need to understand this change better too.

jd-lara (approved):

PR #473, opened 2025-06-23
Add get_num_components

daniel-thom:


jd-lara (approved):
Do we need the companion to this in PSY 5

PR #474, opened 2025-06-24
Add function get_base_component_type

daniel-thom:
This new function will allow parent types to use all default methods in ComponentContainer.


src/component_container.jl
```diff
@@ -22,6 +22,9 @@ abstract type ComponentContainer <: InfrastructureSystemsContainer end
 # The notion of availability used in `get_available` and `get_available_component(s)` is up
 # to the subtype, but it must be consistent across the three functions.
 
+"Return the base type stored in the container."
+get_base_component_type(::ComponentContainer) = InfrastructureSystemsComponent
```
GabrielKS, line 26: Add to docstring and developer-oriented comment above

GabrielKS (changes requested):

PR #477, opened 2025-07-15
Refactor supplemental attributes api

daniel-thom:
This PR simplifies the supplemental attribute query API. Secondarily, it adds the StringTemplates library to build SQL strings.

### DIFF
--- a/src/supplemental_attribute_associations.jl
+++ b/src/supplemental_attribute_associations.jl
@@ -139,16 +139,6 @@ end
 
 const _QUERY_DROP_ATTR_TABLE = "DROP TABLE IF EXISTS $SUPPLEMENTAL_ATTRIBUTE_TABLE_NAME"
 
-"""
-Drop the supplemental attribute associations table.
-"""
-function drop_table(associations::SupplementalAttributeAssociations)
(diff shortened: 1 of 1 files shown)


src/system_data.jl
```diff
@@ -1072,45 +1072,34 @@ end
...
 - `attribute::SupplementalAttribute`: Only return components associated with this attribute.
+- `component_type::Union{Nothing, Type{<:InfrastructureSystemsComponent}}`: Optional, type of the
+  components to return. Can be concrete or abstract.
 """
 function get_associated_components(
-    component_type::Type{<:InfrastructureSystemsComponent},
     data::SystemData,
-    attribute::SupplementalAttribute,
+    attribute::SupplementalAttribute;
+    component_type::Union{Nothing, Type{<:InfrastructureSystemsComponent}} = nothing,
```
daniel-thom, line 1102: Note this change. I did mention it yesterday, but I'm not sure it completely registered with you. I think this is the best interface, but it is different than the `get_components` interface. `component_type` is an optional argument placed last. There is not a separate method with component_type listed first.

I think this is ok because `get_components` requires the component type argument. But, there could be a debate here.

jd-lara (approved):

PR #478, opened 2025-07-21
Do not merge: Convert SQLite time series resolutions from infrasys

daniel-thom:
This will hopefully be temporary. @pesap Can we change infrasys to match the behavior of InfrastructureSystems.jl?

### DIFF
--- a/src/time_series_metadata_store.jl
+++ b/src/time_series_metadata_store.jl
@@ -81,6 +81,8 @@ function _process_migrations_if_needed(store::TimeSeriesMetadataStore)
         _migrate_from_v2_3(store)
     elseif _needs_migration_from_v2_4(store.db)
         _migrate_from_v2_4(store)
+    else
+        _handle_infrasys_resolutions(store)
     end
 end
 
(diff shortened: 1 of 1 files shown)


src/time_series_metadata_store.jl
```diff
@@ -275,6 +277,50 @@ function _migrate_from_v2_4(store::TimeSeriesMetadataStore)
...
+P0DT1H for 1 hour and P0DT5M for 5 minutes.
+InfrastructureSystems.jl always stores regular resolutions in milliseconds.
+InfrastructureSystems.jl does this because user queries might include resolution and we don't
+want to have to search for any conceivable resolution.
+This will go away if we change infrasys to match.
+"""
+function _handle_infrasys_resolutions(store::TimeSeriesMetadataStore)
+    hours_regex = r"^P0DT\d+H$"
+    minutes_regex = r"^P0DT\d+M$"
+    seconds_regex = r"^P0DT\d+S$"
```
daniel-thom, line 291: This doesn't handle days or weeks. If this solution is temporary, no big deal. If it is to be permanent, we should expand this logic or implement a different solution.

PR #479, opened 2025-07-21
Add restriction on deletion of SingleTimeSeries

daniel-thom:
Prevent users from deleting a SingleTimeSeries when it is attached to a DeterministicSingleTimeSeries. Reason: it causes unnecessary complexity with serialization/de-serialization to/from the SiennaGridDB.


jd-lara (approved):

jd-lara:
@daniel-thom the test `test_hybrid_system` broke in PSY due to this change. FYI there is another broken test on the PSSe parsing that's expected to fail if you run all the test suite for PSY

PR #482, opened 2025-09-02
Return results in long format

daniel-thom:


src/Optimization/optimization_problem_results.jl
```diff
@@ -305,55 +305,87 @@ function _copy_for_serialization(res::OptimizationProblemResults)
...
+        num_components = DataFrames.nrow(unique(df[:, component_cols]))
+        num_rows = DataFrames.nrow(df)
+        if num_rows % num_components != 0
+            error(
+                "num_rows = $num_rows is not divisible by num_components = $num_components",
+            )
+        end
+        num_rows_per_component = num_rows ÷ num_components
+        if num_rows_per_component == 1 && num_rows_per_component < length(time_ids)
+            # Case 1: Can't add a time column. TODO: Explain why this happens.
```
daniel-thom, line 345: The number of rows is less than the number of timestamps. Can happen when one variable in the model has 5-minute resolution and another variable has 1-hour resolution.

src/Optimization/optimization_problem_results.jl
```diff
@@ -305,55 +305,87 @@ function _copy_for_serialization(res::OptimizationProblemResults)
...
+    results = Dict{OptimizationContainerKey, DataFrame}()
+    IS.@assert_op time_ids == collect(1:length(timestamps))
+    df_timestamps = DataFrame(:DateTime => timestamps, :time_index => time_ids)
+
+    for (key, df) in result_values
+        if !in(key, container_keys)
+            continue
+        end
+        component_cols = [:component]
+        if "component_x" in names(df)  # TODO DT: what should this name be?
```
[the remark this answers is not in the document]
daniel-thom: Jose and I agree on `:name2`.

src/Optimization/optimization_problem_results.jl
```diff
@@ -305,55 +305,78 @@ function _copy_for_serialization(res::OptimizationProblemResults)
...
+    results = Dict{OptimizationContainerKey, DataFrame}()
+    IS.@assert_op time_ids == collect(1:length(timestamps))
+    df_timestamps = DataFrame(:DateTime => timestamps, :time_index => time_ids)
+
+    for (key, df) in result_values
+        if !in(key, container_keys)
+            continue
+        end
+        component_cols = [:name]
+        if "name2" in names(df)
```
GabrielKS, line 328: Would be nice to factor `name` and `name2` out into constants

src/Optimization/optimization_problem_results.jl
```diff
@@ -305,55 +305,78 @@ function _copy_for_serialization(res::OptimizationProblemResults)
...
+            push!(component_cols, ":name2")
```
GabrielKS, line 329: Is this untested? Seems like it should throw ```ERROR: MethodError: Cannot `convert` an object of type String to an object of type Symbol```

GabrielKS (changes requested):
Some initial comments before I look at the PSI companion.

src/Optimization/optimization_problem_results.jl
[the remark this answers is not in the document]
daniel-thom: In PSI the second column can sometimes represent the system. My only other idea was tying the column names to the OptimizationContainerKey. I'm not sure if that could be reliable.

I'm not even sure I like all of this code being in IS. I'm open to ideas. I don't really like `name2` and don't consider it final-final.

test/test_optimization_results.jl
```diff
@@ -86,24 +99,40 @@ const IS = InfrastructureSystems
         mktempdir(),
         mktempdir(),
     )
-    # Check that variable has time series
+
+    # Test with convert_result_to_natural_units = false
+    @eval IS.Optimization.convert_result_to_natural_units(::Type{<:MockVariable}) = false
```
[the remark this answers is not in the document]
daniel-thom: Seems that I forgot to delete it from the old code.

src/Optimization/optimization_problem_results.jl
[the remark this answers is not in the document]
GabrielKS: True, so maybe PSI sets second column to `component` when it's components and `system` (or no second index column? really those are 1D results, right?) when it's system? Really just hoping for something less generic than `name` and `name2`…

src/Optimization/optimization_problem_results.jl
daniel-thom, line 328: This code block needs a test.

src/enums.jl
jd-lara, line 1: nit: We have been naming the file with enums and constant, definitions.jl in the other repos. This will make it easier to find wrt to the other packages.

daniel-thom: I looked for a file of that name, but it didn't exist. There was this [file](https://github.com/NREL-Sienna/InfrastructureSystems.jl/blob/main/src/Optimization/enums.jl), so I followed its convention. I'm still ok renaming it.

src/Optimization/abstract_model_store.jl
```diff
@@ -30,10 +30,6 @@ abstract type AbstractModelStore end
 # - :expressions
 
 # Required methods for subtypes:
-# - initialize_storage!
```
daniel-thom, line 33: There is something I found confusing but did not try to fix. PSI imports IS.Optimization.read_optimizer_stats but not these others. So, implements its own version of that method but then creates PSI methods for the others. There is not a clean separation of what is PSI and what is IS.Optimization.

src/Optimization/optimization_problem_results.jl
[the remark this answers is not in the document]
daniel-thom: There is now a mechanism to override the column names based on the OptimizationContainerKey.

jd-lara (approved):

PR #484, opened 2025-09-23
include name in output of show_time_series_data

luke-kiernan:


daniel-thom (approved):
Thanks. I'm surprised that we didn't already have this.

Issue #485, opened 2025-09-30
Implement Method to get all the outages instances attached to a component type


SebastianManriqueM:
For the security constrained formulations in PSY 5 (N-1 and G-1) I need obtain all the outages instances that have been associated with a given component type. It worths to note that the user can attach the same outage instance to 2 or more different components (To simulate G-1, G-1. G-3...etc), so when It is used "get_component_supplemental_attribute_pairs()" I might have the same supplemental attribute more than once in the pairs.

Currently I'm doing this in the following way:

`
generator_outages_pairs = PSY.get_component_supplemental_attribute_pairs(
        PSY.Generator,
        PSY.UnplannedOutage,
        sys,
    )
    associated_outages = unique([outage for (_, outage) in generator_outages_pairs])
`

According to this documentation: https://github.com/NREL-Sienna/PowerSystems.jl/blob/psy5/docs/src/explanation/supplemental_attributes.md, It was the best way that I found to implement that.

- How to get all the outages instances of a specific outage type associated to a component instance - get_supplemental_attributes()
- How to get all the component instances associated with an specific outage instance - get_associated_components()
- How to the components associated with an attribute type, **but It is not available the dual method which is what I need ( Get the outages associated with a component type)**

daniel-thom:
I'm not opposed to making a new function. However, do we really need it? The existing solution is not that bad, right? You can filter on component and attribute types; you just have to call `unique` on the result yourself.

My only issue with making the new function is that I sense that there will continue to be more and more of these for lots of different use cases.

I'll mention a generic solution we've discussed previously. Suppose we make one new function on top of `get_component_supplemental_attribute_pairs` (which includes filters) that returns a DataFrame like this:
```
8×2 DataFrame
 Row │ component                          attribute
     │ Component                          Suppleme…
─────┼──────────────────────────────────────────────────────────────────────
   1 │ ThermalStandard(Solitude, true, …  GeographicInfo(Dict{String, Any}…
   2 │ ACBus(3, nodeC, true, ACBusTypes…  GeographicInfo(Dict{String, Any}…
   3 │ ThermalStandard(Park City, true,…  GeographicInfo(Dict{String, Any}…
   4 │ ACBus(1, nodeA, true, ACBusTypes…  GeographicInfo(Dict{String, Any}…
   5 │ ThermalStandard(Solitude, true, …  GeometricDistributionForcedOutag…
[...]

PR #489, opened 2025-10-09
get_supplemental_attribute methods with component type input

luke-kiernan:
Address issue #485.

I also restructured things so that the code structure for `get_associated_components` on an attribute and `get_associated_supplemental_attributes` on a component are near-identical. (This accounts for the removals around 470-500, and the additions from 300-400.) If those things are undesired, I can put them back.

### DIFF
--- a/test/test_function_data.jl
+++ b/test/test_function_data.jl
@@ -209,10 +209,11 @@ end
 end
 
 @testset "Test FunctionData printing" begin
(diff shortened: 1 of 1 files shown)


jd-lara:
Looks like there is one more test pending that is also related to Julia 1.12 printing of Named Tuples 

image

@luke-kiernan can you fix this one too please

PR #505, opened 2025-11-07
Base.iterate for TypeSortedCollections

luke-kiernan:
Yet to do profiling, but it's working. Tests were mostly written by CoPilot.

The tests that are failing are unrelated, would be fixed by rebasing onto main. [I tried doing that, but got conflicts.]


luke-kiernan:
The problem still persists, but the solution in this PR has fallen out of favor. Explicitly caching the collection in a type-stable way (see PNM's [YBusACBranches](https://github.com/Sienna-Platform/PowerNetworkMatrices.jl/blob/main/src/YbusACBranches.jl)) seems like a better approach.

PR #509, opened 2025-11-11
Review codebase for duplication and optimization

jd-lara:


src/Optimization/optimization_problem_results.jl
```diff
@@ -123,7 +123,7 @@ end
 function export_result(
     ::Type{CSV.File},
     path,
-    name::AbstractString,
+    name::String,
```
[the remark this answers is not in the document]
GabrielKS: agree that this is problematic

src/time_series_utils.jl
```diff
@@ -11,6 +11,26 @@ const TIME_SERIES_STRING_TO_TYPE = Dict(
     "SingleTimeSeries" => SingleTimeSeries,
 )
 
+"""
+Parse time series type string to concrete type using static dispatch.
+This is more efficient for precompilation than dictionary lookup.
```
GabrielKS: It… replaced a dictionary lookup with a long if/else chain. That is not a style we want to encourage! If we're really concerned about allocations here, we might prototype whether a `Val` dispatch approach is better, but definitely not this.

src/utils/logging.jl
```diff
@@ -60,22 +60,24 @@ end
...
 """
 function report_log_summary(tracker::LogEventTracker)
-    text = "\nLog message summary:\n"
+    io = IOBuffer()
+    print(io, "\nLog message summary:")
     # Order by criticality.
     for level in sort!(collect(keys(tracker.events)); rev = true)
         num_events = length(tracker.events[level])
-        text *= "\n$num_events $level events:\n"
+        print(io, "\n\n", num_events, " ", level, " events:")
```
GabrielKS: Like, I _guess_ this is slightly more performant, and it's sylistically inoffensive, but the context is so very much not performance critical here.

src/utils/print.jl
```diff
@@ -305,7 +305,7 @@ function show_supplemental_attributes(component::InfrastructureSystemsComponent)
 end
 
 function show_supplemental_attributes(io::IO, component::InfrastructureSystemsComponent)
-    data_by_type = Dict{Any, Vector{OrderedDict{String, Any}}}()
+    data_by_type = Dict{DataType, Vector{OrderedDict{String, Any}}}()
```
GabrielKS: I approve this change

PR #512, opened 2025-11-17
Create PiecewiseLinearData conversion functions

jd-lara:
…pData

Add bidirectional conversion capabilities:
- PiecewiseStepData(::PiecewiseLinearData): converts by computing slopes (derivative)
- PiecewiseLinearData(::PiecewiseStepData, initial_y): converts by running sum (integral)
- Corresponding Base.convert methods for both directions

The initial_y parameter allows preserving the y-offset when converting back from step data to linear data.

### DIFF
--- a/test/test_function_data.jl
+++ b/test/test_function_data.jl
@@ -612,6 +612,70 @@ end
     end
 end
 
+@testset "Test PiecewiseLinearData <-> PiecewiseStepData constructor conversions" begin
+    # Test basic conversion: PiecewiseLinearData -> PiecewiseStepData
+    linear_data = IS.PiecewiseLinearData([(0.0, 0.0), (1.0, 2.0), (3.0, 5.0)])
+    step_data = IS.PiecewiseStepData(linear_data)
+    @test IS.get_x_coords(step_data) == [0.0, 1.0, 3.0]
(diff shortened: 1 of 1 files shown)


jd-lara:
@GabrielKS I am trying to figure out a reasonable set of  utilities to handle the need for conversions mostly for internal handling in PSI. I don't think we should do that inside of the modeling library and given the need to have some utility to clean datasets I am revising my position.

GabrielKS:
> @GabrielKS I am trying to figure out a reasonable set of utilities to handle the need for conversions mostly for internal handling in PSI. I don't think we should do that inside of the modeling library and given the need to have some utility to clean datasets I am revising my position.

That makes sense — I think we should still probably do this only at the `ValueCurve` level. If your `FunctionData` fundamentally represents an `InputOutputCurve`, ideally it would be represented as an `InputOutputCurve` all along, but in cases (e.g., time series :( ) where that's not possible, I'd recommend upgrading to `ValueCurve` on the fly to take advantage of these idiomatic conversions that automatically handle the initial input, etc.:

a0717260-40ff-40d4-a9d5-4c1654573201

jd-lara:
closing after @GabrielKS comments.

PR #514, opened 2025-11-20
implement evaluation of piecewise curves

luke-kiernan:
Evaluation as a function, for `PiecewiseIncrementalCurve` and `PiecewiseAverageCurve`. I also added `check_domain(fd::FunctionData, x::Real)`, as defense against floating point errors.

### DIFF
--- a/src/cost_aliases.jl
+++ b/src/cost_aliases.jl
@@ -45,6 +45,8 @@ Base.show(io::IO, vc::LinearCurve) =
         Base.show_default(io, vc)
     end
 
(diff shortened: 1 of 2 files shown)


luke-kiernan:
We've agreed that it makes sense to evaluate an `InputOutputCurve` at an input `x`: just evaluate the function data at `x`. However, it's ambiguous for the `Incremental` and `AverageRate` curves, so there, I'll leave evaluation un-implemented. (If you want to evaluate the curve--not the rate--for one of those, just convert it to an `InputOutputCurve` first.)

Will do in a separate PR.

PR #516, opened 2025-11-21
Joint dependency updates and Claude improvement recommendations 

jd-lara:
This PR follows the same approach that DataFrames.jl had for backwards compatibility of PrettyTables and avoid a major release


Project.toml
```diff
@@ -47,7 +47,7 @@ JSON3 = "^1.11"
 Logging = "1"
 Mustache = "1"
 Pkg = "1"
-PrettyTables = "1.3 - 3"
+PrettyTables = "2.1, 3.1"
```
hannahchubin: @jd-lara is this how I should implement the PrettyTables version for all the repos, or is this a special case for IS?

PR #517, opened 2025-11-21
InputOutputCurve function evaluation

luke-kiernan:
Implement `(ioc::InputOutputCurve)(x::Real)` by evaluating `ioc`'s function data at `x`. (This gives me a clean way to calculate the power loss in a HVDC line in `PowerFlows.jl`.) Also fix a copy-paste error and only enforce bounds up to FP precision.


GabrielKS (approved):
LGTM, thanks!

PR #526, opened 2025-12-23
Claude/convex approximation method 01 n xc pz4 xpj2 ys w4 f uur n8 x9

PabloBotinGP:
During my work validating operational cost curves of `ThermalStandard` generators (see https://github.nrel.gov/SIIP/EasternInterconnection.jl/issues/86), I used and extended several methods in this branch to analyze and convexify non-convex functions.

This PR includes three main contributions:

1. **Stricter convexity / concavity checks**  
   Modified `is_convex` and `is_concave` to return `false` when the function is not strictly convex/concave, explicitly flagging linear (constant-slope) functions as invalid. This ended up not being really used by my scripts but wanted to correct that. 

2. **New non-convexity check**  
   Added `is_nonconvex()`, which provides a clear and reusable way to identify non-convex cost curves and simplifies downstream QA logic.

3. **Analysis and example scripts**  
   Included scripts that use the new methods (including PAVA convexification and approximation-error analysis). These scripts serve both as tests and as usage examples, and include tools for generating summary statistics. They are designed to work with all PSY cost-curve types, but have so far been tested only with piecewise-linear curves.

### DIFF
--- a/scripts/Convexity_Check.jl
+++ b/scripts/Convexity_Check.jl
@@ -1,0 +1,148 @@
+ # Non-convexity check script for generator cost curves
+# Outputs: console summary + non_convex_generators.json
+
+using PowerSystems
+using InfrastructureSystems
+const PSY = PowerSystems
+const IS = InfrastructureSystems
+# System path - edit as needed
+const sys_json_path = "/path/to/your/system.json"
+
+function check_nonconvex(curve)
+    # Check if a curve is non-convex using IS.is_nonconvex. Returns a boolean.
+    try
+        if isa(curve, InfrastructureSystems.PiecewiseIncrementalCurve)
+            # Extract the underlying piecewise data and check for non-convexity
+            function_data = PSY.get_function_data(curve)
+            return IS.is_nonconvex(function_data)
+        elseif isa(curve, PSY.PiecewiseLinearCurve)
+            # Extract the underlying piecewise data and check for non-convexity
+            function_data = PSY.get_function_data(curve)
+            return IS.is_nonconvex(function_data)
+        elseif isa(curve, PSY.QuadraticCurve)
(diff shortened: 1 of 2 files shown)


scripts/Convexity_Check.jl
```diff
@@ -0,0 +1,148 @@
...
+using InfrastructureSystems
+const PSY = PowerSystems
+const IS = InfrastructureSystems
+# System path - edit as needed
+const sys_json_path = "/path/to/your/system.json"
+
+function check_nonconvex(curve)
+    # Check if a curve is non-convex using IS.is_nonconvex. Returns a boolean.
+    try
+        if isa(curve, InfrastructureSystems.PiecewiseIncrementalCurve)
```
GabrielKS: Use multiple dispatch. Different methods for different types:

```julia
function check_nonconvex(curve::InfrastructureSystems.PiecewiseIncrementalCurve)
  # body 1
end

function check_nonconvex(curve::PSY.PiecewiseLinearCurve)
  # body 2
end

# etc.
```

src/function_data.jl
```diff
@@ -357,9 +357,39 @@ is_convex(pwl::PiecewiseStepData) =
...
+    is_concave(data::FunctionData) -> Bool
+
+Returns `true` if the function data represents a concave function, `false` otherwise.
+
+- `LinearFunctionData`: Always returns `true` (linear functions are both convex and concave)
+- `QuadraticFunctionData`: Returns `true` if quadratic_term ≤ 0
+- `PiecewiseLinearData`: Returns `true` if slopes are non-increasing
+- `PiecewiseStepData`: Returns `true` if y-coordinates are non-increasing
 """
+is_concave(::LinearFunctionData) = true
```
GabrielKS: this seems okay I think

PR #527, opened 2026-01-01
fixes for 1.13

ymiftah:
Addresses https://github.com/NREL-Sienna/PowerSystems.jl/issues/1600. 

Some fixes required due to changes in the way iterators work in 1.13. Changes proposed should be backward compatible.
1. Adjust the `Base.iterate` specialisation to the new signatures (with/without state)
2. Adjust the ingestion of time series from DataFrames / Tables

### DIFF
--- a/src/utils/flatten_iterator_wrapper.jl
+++ b/src/utils/flatten_iterator_wrapper.jl
@@ -14,9 +14,15 @@ end
 
 Base.@propagate_inbounds function Base.iterate(
     iter::FlattenIteratorWrapper{T, I},
-    state = (),
 ) where {T, I}
-    state === () ? Base.iterate(iter.iter) : Base.iterate(iter.iter, state)
+    Base.iterate(iter.iter)
+end
+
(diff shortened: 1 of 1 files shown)


src/single_time_series.jl
```diff
@@ -110,6 +110,10 @@ function SingleTimeSeries(
     else
         error("fatal: $(typeof(data))")
     end
+    # TimeArray's table integration (correctly) returns a Matrix as values, even if size in column dimension is 1 (julia +1.13)
+    # As the rest expects a single valued timeseries, we slice to the only columns available to obtain the appropriate Vector value
```
ymiftah: I have not checked across other use cases / dependents: from the PowerSystems / PowerSimulaions docs, it seems that what is described above is the typical use case: only a single value column other than the timestamp,

Is this implicit select a possible issue ? Should the `throw` be replaced by a warning ?

daniel-thom (approved):

PR #535, opened 2026-01-15
add retest migration guide

luke-kiernan:


jd-lara (commented):
Can you make some of this instructions for Sienna specific like how to load artifacts from the TestData and the use of the IS logging please.

daniel-thom (approved):

luke-kiernan:
> Can you make some of this instructions for Sienna specific like how to load artifacts from the TestData and the use of the IS logging please.

I added another page on how to use your local copy of `PowerSystemsTestData` instead of the Julia official artifacts. For logging, what are you looking for that isn't addressed in the current logging page? Just "here's how to send a log message with a specific IS logging group?"

PR #536, opened 2026-01-16
add some improvements to SQL queries

jd-lara:
This is a simplified version after testing of which SQL improvements actually make a dent

### DIFF
--- a/src/time_series_metadata_store.jl
+++ b/src/time_series_metadata_store.jl
@@ -777,7 +777,7 @@ end
 Return the number of unique time series arrays.
 """
 function get_num_time_series(store::TimeSeriesMetadataStore)
-    return _execute_count_direct(
+    return _execute_count(
         store,
         "SELECT COUNT(DISTINCT time_series_uuid) AS count FROM $ASSOCIATIONS_TABLE_NAME",
     )
@@ -808,10 +808,10 @@ function get_time_series_counts(store::TimeSeriesMetadataStore)
         WHERE interval IS NOT NULL
     """
 
-    count_components = _execute_count_direct(store, query_components)
-    count_attributes = _execute_count_direct(store, query_attributes)
-    count_sts = _execute_count_direct(store, query_sts)
-    count_forecasts = _execute_count_direct(store, query_forecasts)
+    count_components = _execute_count(store, query_components)
+    count_attributes = _execute_count(store, query_attributes)
+    count_sts = _execute_count(store, query_sts)
+    count_forecasts = _execute_count(store, query_forecasts)
 
     return TimeSeriesCounts(;
         components_with_time_series = count_components,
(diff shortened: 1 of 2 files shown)


src/time_series_metadata_store.jl
```diff
@@ -442,6 +449,35 @@ function _create_indexes!(store::TimeSeriesMetadataStore)
         unique = false,
         ifnotexists = true,
     )
+    # New indexes for improved query performance
+    SQLite.createindex!(
```
daniel-thom: Can you explain why we should add these extra indexes? There is a memory cost and I do not think they are used in performance-sensitive areas.

src/time_series_metadata_store.jl
```diff
@@ -561,6 +600,19 @@ function clear_metadata!(store::TimeSeriesMetadataStore)
     _execute(store, "DELETE FROM $ASSOCIATIONS_TABLE_NAME")
 end
 
+"""
+Update database statistics for optimal query planning.
+Call this after bulk operations or significant data changes.
+"""
+function optimize_database!(store::TimeSeriesMetadataStore)
```
daniel-thom: As with the other method, this is not called.

src/time_series_metadata_store.jl
```diff
@@ -1306,17 +1352,14 @@ end
...
 const _QUERY_CHECK_FOR_ATTACHED_DSTS = """
 SELECT time_series_uuid
 FROM $ASSOCIATIONS_TABLE_NAME
 WHERE time_series_uuid = ?
 GROUP BY time_series_uuid
 HAVING
-    SUM(time_series_type = 'SingleTimeSeries') = 1
-    AND
-    SUM(time_series_type = 'DeterministicSingleTimeSeries') >= 1;
+    SUM(CASE WHEN time_series_type = 'SingleTimeSeries' THEN 1 ELSE 0 END) = 1
```
daniel-thom: When we wrote the original query I tried several variations based LLM recommendations, including something like this. I spent some time profiling them in various real-world situations. I picked the one with the best performance. Can you post results showing that this new query is better, and under what scenarios?

src/time_series_metadata_store.jl
```diff
@@ -1573,16 +1628,23 @@ function _make_category_clause(ts_type::Type{<:TimeSeriesData})
     return clause, subtypes
 end
 
+# Multiple dispatch helpers for feature pattern generation
+# Pattern for string values: match key-value pair in JSON structure
+_make_feature_pattern(key::String, val::AbstractString) = "%\"$(key)\":\"$(val)\"%"
+
+# Pattern for non-string values (Bool, Int): match key-value pair in JSON structure
+_make_feature_pattern(key::String, val::Union{Bool, Int}) = "%\"$(key)\":$(val)%"
```
daniel-thom: We support float values in features, so this needs to be changed.

PR #545, opened 2026-01-27
Convexification of cost curves

PabloBotinGP:
## Description

Main implementation: 

- **`make_function(...)`**  
which uses:
- **`isotonic_regression()`** (already defined before)
- **`is_convex(...)`** (reviewed and defined multiple dispatch)

It relies on multiple dispatch so it can handle:

### ValueCurve types
- `InputOutput`
- `Incremental`
- `AverageRate`

### Data structures
- `Linear`
- `Quadratic`
- `PiecewisePoint`
- `PiecewiseIncremental`
- `PiecewiseAverage`

This covers the full set of curve/data representations currently used in Sienna.

These aliases are useful for understanding the CurveTypes: 
- LinearCurve = InputOutputCurve{LinearFunctionData}
- QuadraticCurve = InputOutputCurve{QuadraticFunctionData}
- PiecewisePointCurve = InputOutputCurve{PiecewiseLinearData}
- PiecewiseIncrementalCurve = IncrementalCurve{PiecewiseStepData}
- PiecewiseAverageCurve = AverageRateCurve{PiecewiseStepData}

---

### Linear
Always convex by definition, nothing to do here.

### Quadratic
Handled in a very simple way for now:  
The quadratic term is removed and the function is reduced to a linear one.

This is not the best solution, but:
- I presume these curves are very rare in practice, based on limited experience, tho.
- doing it “properly” would require changing how `isotonic_regression()` works

This should be revisited in the future if quadratics become relevant and we consider this approximation is too vague. 

### PiecewisePoint
Steps:
1. Convert points → slopes  
2. Convexify slopes  
3. Reconstruct the curve as points

### PiecewiseIncremental
Straightforward: convexification is applied directly in slope space.  
This is the most common structure (at least in EI), so it’s good that this path is simple and robust.

### PiecewiseAverage
Handled by converting to incremental form and using the same logic.

---

### InputOutput
Direct convexification.

### Incremental
1. Integrate to total cost  
2. Convexify  
3. Integrate back

### AverageRate
1. Integrate to total cost  
2. Convexify  
3. Integrate back

This keeps the behavior consistent across representations.

---

## File structure

Right now everything lives in:

- `make_convex.jl`
- `convexity_checks.jl`

This was done for clarity and review.

---

## Testing

- New tests added for the newly implemented functions — all passing  
- Full `InfrastructureSystems` test passes  
- Toy-curve tests pass  
- Tests with real models are in progress  
  - Sara is testing with **EST**  
  - Sara will generate plots of original vs. convexified curves  
- If we get access to a large model with many non-convex curves, we’re happy to:
  - run convexification at scale  
  - generate statistical summaries  
  - produce distributions/plots to characterize the approximation error

---

### 1) File placement
Currently the logic is in `make_convex.jl` and `convexity_checks.jl`.

My understanding is that after review this should probably live in:

- **`function_data.jl`**

since these are methods over `FunctionData` types.  
Confirming that this is the intended structure would be helpful.

---

### 2) Multiple dispatch
There are a lot of method definitions.

It would be good to review whether any of them are redundant and can be safely removed.

---

### 3) Documentation
It would be good to properly define and document these functions in the package docs.

Is this something we need to investigate/learn/do?

@sabrilg-nrel  Thanks for all the work and feel free to add any more info you would like to add and update the results in form of comments on this PR.

### DIFF
--- a/test/test_function_data.jl
+++ b/test/test_function_data.jl
@@ -1040,64 +1040,6 @@ end
     @test IS.get_function_data(convex2) == IS.get_function_data(convex1)
 end
 
-@testset "Test make_convex for LinearFunctionData (via ValueCurve)" begin
-    # LinearFunctionData is always convex, test via InputOutputCurve
-    lfd = IS.LinearFunctionData(5.0, 1.0)
-    @test IS.is_convex(lfd)
-    curve = IS.InputOutputCurve(lfd)
-    result = IS.make_convex(curve)
-    @test result === curve  # Same object returned
-
-    # Test with negative slope (still convex - linear is both convex and concave)
-    lfd_neg = IS.LinearFunctionData(-3.0, 10.0)
-    @test IS.is_convex(lfd_neg)
-    curve_neg = IS.InputOutputCurve(lfd_neg)
-    result_neg = IS.make_convex(curve_neg)
-    @test result_neg === curve_neg
-
-    # Test with zero slope
-    lfd_zero = IS.LinearFunctionData(0.0, 5.0)
-    @test IS.is_convex(lfd_zero)
-    curve_zero = IS.InputOutputCurve(lfd_zero)
-    result_zero = IS.make_convex(curve_zero)
-    @test result_zero === curve_zero
-end
-
-@testset "Test make_convex for QuadraticFunctionData (via ValueCurve)" begin
-    # Test convex quadratic (a > 0) - should return same object
-    qfd_convex = IS.QuadraticFunctionData(2.0, 3.0, 4.0)
-    @test IS.is_convex(qfd_convex)
-    curve = IS.InputOutputCurve(qfd_convex)
-    result = IS.make_convex(curve)
-    @test result === curve
-
-    # Test linear quadratic (a = 0) - should return same object
-    qfd_linear = IS.QuadraticFunctionData(0.0, 3.0, 4.0)
-    @test IS.is_convex(qfd_linear)
-    curve_linear = IS.InputOutputCurve(qfd_linear)
-    result_linear = IS.make_convex(curve_linear)
-    @test result_linear === curve_linear
-
-    # Test concave quadratic (a < 0) with infinite domain throws error
-    # QuadraticFunctionData has infinite domain (-Inf, Inf) by default
-    qfd_concave = IS.QuadraticFunctionData(-2.0, 3.0, 4.0)
-    @test !IS.is_convex(qfd_concave)
-    curve_concave = IS.InputOutputCurve(qfd_concave)
-    @test_throws ArgumentError IS.make_convex(curve_concave)
-
-    # Test with small negative quadratic term (edge case)
-    qfd_small_neg = IS.QuadraticFunctionData(-1e-12, 5.0, 10.0)
-    curve_small = IS.InputOutputCurve(qfd_small_neg)
-    # Due to tolerance in is_convex, this might be considered convex
-    # If convex, make_convex returns unchanged; if not, throws due to infinite domain
-    if IS.is_convex(curve_small)
-        result_small = IS.make_convex(curve_small)
-        @test IS.is_convex(IS.get_function_data(result_small))
-    else
-        @test_throws ArgumentError IS.make_convex(curve_small)
-    end
-end
-
 @testset "Test convex approximation with random data" begin
     rng = Random.Xoshiro(42)
     n_tests = 20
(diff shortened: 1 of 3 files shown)


PabloBotinGP:
> This figure illustrates the convexification of three curve types: an input–output curve, an incremental (heat-rate) curve, and a quadratic curve. The incremental curve is first integrated and plotted as its corresponding input–output curve before convexification. In all cases, the original non-convex curve is shown alongside its convexified counterpart. Screenshot 2026-01-27 at 11 58 57

This figure is a clear evidence about how the convexification method for quadratic curves is unacceptable. 
Looking at this, it would make more sense to convert these type of curves into a linear curve using the power range edges. 
Will work on that

src/function_data/make_convex.jl
```diff
@@ -0,0 +1,528 @@
...
+
+Merge colinear segments in a `PiecewisePointCurve` (InputOutputCurve{PiecewiseLinearData}).
+
+Algorithm:
+1. Compute slopes between consecutive points
+2. Identify groups of consecutive segments with slopes within tolerance ε
+3. For each colinear group, keep only the first and last points
+4. Preserve first and last endpoints exactly
+"""
+function merge_colinear_segments(
```
jd-lara: Do we need to make a new object or can we modify in place?

src/function_data/make_convex.jl
```diff
@@ -0,0 +1,528 @@
...
+  - `Vector{Float64}`: custom weights per segment
+- `anchor`: Point preservation strategy for reconstructing points from new slopes
+  - `:first` (default): preserve first point, propagate forward
+  - `:last`: preserve last point, propagate backward
+  - `:centroid`: minimize total vertical displacement
+- `merge_colinear`: Whether to merge colinear segments before convexification (default: `true`)
+  - Colinear segments are those where consecutive slopes differ by less than a tolerance
+  - This cleanup step removes artificial segmentation that can cause false non-convex detections
+"""
+function make_convex end
```
jd-lara: we should have an error here for not implemented

src/function_data/make_convex.jl
```diff
@@ -0,0 +1,528 @@
...
+    ymax = fd(Pmax)
+
+    # Build linear function connecting (Pmin, ymin) to (Pmax, ymax)
+    # slope: m = (ymax - ymin) / (Pmax - Pmin)
+    # intercept: b = ymin - m * Pmin
+    m = (ymax - ymin) / (Pmax - Pmin)
+    b = ymin - m * Pmin
+
+    new_fd = LinearFunctionData(m, b)
+    return InputOutputCurve(new_fd, get_input_at_zero(curve))
```
jd-lara: Add an info statement that the function type has been changed

src/function_data/make_convex.jl
```diff
@@ -0,0 +1,528 @@
...
+
+# Arguments
+- `curve`: A `QuadraticCurve` (`InputOutputCurve{QuadraticFunctionData}`)
+
+# Returns
+- If already convex: returns the original curve unchanged
+- If concave: returns an `InputOutputCurve{LinearFunctionData}` (LinearCurve)
+  that connects the endpoints of the original curve
+"""
+function make_convex(curve::InputOutputCurve{QuadraticFunctionData})
```
jd-lara: I don't think this one should be called make convex since it gets one type and returns other type

jd-lara:
@PabloBotinGP let's not try to "convexify" quadratics that's not really a good thing.

src/function_data/make_convex.jl
PabloBotinGP: You mean, rather than disassembling and re-assembling, modify the object in place? 
Let me investigate

src/function_data/make_convex.jl
PabloBotinGP: Ok.

PabloBotinGP:
> I am getting to the opinion that we should change the `make_convex` for a better function name like fix_convexity or something like that @daniel-thom any thoughts?

@jd-lara @daniel-thom 

Some proposals: sanitize_curve!, normalize_curve!, curve_convexity_check!, clean_curve!. 

- "curve". I believe it is important to include the 'curve' term because it works with curves. 
- "!". Looking at Julia documentation, I realized that this is necessary since this function is changing the value of the curve.

src/function_data/make_convex.jl
PabloBotinGP: Some proposals: sanitize_curve!, normalize_curve!, curve_convexity_check!, clean_curve!.

"curve". I believe it is important to include the 'curve' term because it works with curves.
"!". Looking at Julia documentation, I realized that this is necessary since this function is changing the value of the curve.

jd-lara (commented):
I'll merge into my branch to make changes locally before a complete pr

sabrilg-nrel:
# WECC Curve Validation Tests

The functions `is_convex`, `make_convex`, and `merge_colinear_segments` were tested on the WECC dataset.  
All functions executed without errors, confirming that multiple dispatch correctly supports all curve types present in the system.

---
## Summary of Results

All generator cost curves in WECC are **already convex**.  
However, a subset contains **colinear (redundant) segments**.

When a curve is convex, `make_convex` correctly skips convexification and applies colinearity cleanup when enabled:

```julia
if is_convex(curve)
    return merge_colinear ? merge_colinear_segments(curve) : curve
end
```
### Generator Cost Curve Quality Summary
Screenshot 2026-01-29 at 16 51 08


    
## Colinearity corrections
  After convexity is verified, `merge_colinear_segments` removes redundant intermediate points with identical slopes.
This reduces artificial segmentation while preserving the original curve shape and economic meaning.

The figure below shows *five representative WECC generators* where colinear points were successfully removed.
image

---
### Colinearity Reduction Statistics

The figure below shows the distribution of colinearity reduction ratios for generators whose cost curves contained colinear segments.

The reduction ratio measures the fraction of redundant breakpoints removed after cleanup.

- Values greater than zero indicate successful removal of redundant points
- Most generators exhibit reductions between 30% and 40%
- No changes were made to convexity or marginal cost structure
image

PR #558, opened 2026-03-14
Jd/timeseries functiondata

jd-lara:
This is part 1 for fixing https://github.com/NREL-Sienna/PowerSystems.jl/issues/1472 and making it easier to test and reuse the code of MBC and other PWL models 

Some of the comments in this PR by copilot I've addressed in #559


GabrielKS (commented):
A high-level design suggestion before diving deeply into #559 : rather than having duplicate structs for every single FunctionData type (`LinearFunctionData` and `TimeSeriesLinearFunctionData`, etc.), why not have a single parametric struct:

```julia
@kwdef struct TimeSeriesFunctionData{T <: FunctionData}
    time_series_key::TimeSeriesKey
end
```

(`LinearFunctionData` and `TimeSeriesFunctionData{LinearFunctionData}`, etc.)? Seems cleaner while getting you the same thing. As you know I think ultimately _all_ time series should be parameterized this way, but you can do this wrapper for FunctionData without making that commitment.

If this incurs a precompilation performance penalty, I think that penalty should be measured and weighed against the drop in cleanliness from doing this.

luke-kiernan (approved):
CoPilot's comments seem worth addressing, bu otherwise looks fine.

jd-lara:
@luke-kiernan the copilot comments are implemented in #559

jd-lara:
closed in favor of #559

PR #559, opened 2026-03-14
Jd/timevariable function

jd-lara:
This is part 2 of https://github.com/NREL-Sienna/PowerSystems.jl/issues/1472. I'd consider just reusing the value curves and have them take an abstract type but it had impacts in the precompilation. @luke-kiernan maybe using @generated here could help but this was more semantically cleaner even if somewhat repetitive.


src/production_variable_cost_curve.jl
```diff
@@ -17,9 +17,30 @@ get_function_data(cost::ProductionVariableCostCurve) =
 get_initial_input(cost::ProductionVariableCostCurve) =
     get_initial_input(get_value_curve(cost))
 "Calculate the convexity of the underlying data"
+function is_convex(curve::ValueCurve{T}) where {T <: TimeSeriesFunctionData}
+    throw(
+        ArgumentError(
+            "Convexity is not defined for time-series-backed ValueCurve; use time-series specific analysis instead.",
```
luke-kiernan, line 23: > use time-series specific analysis instead

Huh? What does that mean?

In order to have uniform code paths downstream, I'd consider introducing `is_convex(curve, t)`, which is well-defined in both cases.

luke-kiernan (approved):
Looks fine. Feels a little clunky having two of each (e.g. `IncrementalCurve` and `TimeSeriesIncrementalCurve`), but if it allows for more uniform interface, then it's worth it.

src/function_data/function_data.jl
```diff
@@ -1,12 +1,33 @@
...
+Data for a linear function: `f(x) = proportional_term * x + constant_term`.
+
+Use this when the output changes at a constant rate with the input — e.g., constant
+marginal cost regardless of production level.
 
 # Arguments
- - `proportional_term::Float64`: the proportional term in the represented function
- - `constant_term::Float64`: the constant term in the represented function
+- `proportional_term::Float64`: slope of the function (e.g., \$/MWh)
+- `constant_term::Float64`: intercept (e.g., no-load cost in \$/h). Defaults to `0.0`.
```
GabrielKS: Pragmatically you might not care but it seems strictly speaking incorrect to refer to units here as if it was a `CostCurve{InputOutputCurve{LinearFunctionData}}`. All `LinearFunctionData` _is_ is just a linear function, it could be _any_ linear function. It doesn't seem like _too_ heavy a lift to tell a power systems person "that struct just represents any y=mx+b function"

src/function_data/function_data.jl
```diff
@@ -47,14 +68,15 @@ function Base.show(io::IO, ::MIME"text/plain", fd::LinearFunctionData)
 end
 
 """
-Structure to represent the underlying data of quadratic functions. Principally used for the
-representation of cost functions
-`f(x) = quadratic_term*x^2 + proportional_term*x + constant_term`.
+Data for a quadratic function: `f(x) = quadratic_term * x^2 + proportional_term * x + constant_term`.
+
+Use this when you have a smooth, continuously increasing marginal cost curve — commonly
+fitted from heat rate tables. Marginal cost increases linearly with output.
```
GabrielKS: This is also strictly speaking incorrect. Nothing about the definition of the struct requires increasing.

src/function_data/function_data.jl
```diff
@@ -108,13 +130,30 @@ function Base.show(io::IO, ::MIME"text/plain", fd::QuadraticFunctionData)
 end
 
 """
-Structure to represent piecewise linear data as a series of points: two points define one
-segment, three points define two segments, etc. The curve starts at the first point given,
-not the origin. Principally used for the representation of cost functions where the points
-store quantities (x, y), such as (MW, \$/h).
+Data for a piecewise linear function defined by (x, y) **value** points.
+
+Each point stores an absolute output: e.g., `(100.0 MW, 500.0 \$/h)`. The function
```
GabrielKS: see above

src/function_data/function_data.jl
```diff
@@ -108,13 +130,30 @@ function Base.show(io::IO, ::MIME"text/plain", fd::QuadraticFunctionData)
...
+Two points define one segment, three define two, etc. The curve starts at the first
+point, not the origin.
+
+If your data gives marginal rates (slopes) between breakpoints rather than total values
+at each point, use [`PiecewiseStepData`](@ref) instead.
+
+**Optimization formulation:** this data shape maps to the **lambda (convex combination)**
+formulation. One λ variable is created per breakpoint (n points → n variables):
+
+    P = Σᵢ λᵢ · Pᵢ,   C = Σᵢ λᵢ · Cᵢ,   Σᵢ λᵢ = on_status,   λᵢ ∈ [0, 1]
```
GabrielKS: use LaTeX?

src/function_data/function_data.jl
```diff
@@ -108,13 +130,30 @@ function Base.show(io::IO, ::MIME"text/plain", fd::QuadraticFunctionData)
...
+
+Each point stores an absolute output: e.g., `(100.0 MW, 500.0 \$/h)`. The function
+linearly interpolates between consecutive points. **The y-values are totals, not slopes.**
+Two points define one segment, three define two, etc. The curve starts at the first
+point, not the origin.
+
+If your data gives marginal rates (slopes) between breakpoints rather than total values
+at each point, use [`PiecewiseStepData`](@ref) instead.
+
+**Optimization formulation:** this data shape maps to the **lambda (convex combination)**
```
GabrielKS: this seems like it breaks an important abstraction boundary, don't we want "represent the data Correctly" to be independent of how it's gonna be formulated in the optimization problem?

src/function_data/function_data.jl
```diff
@@ -108,13 +130,30 @@ function Base.show(io::IO, ::MIME"text/plain", fd::QuadraticFunctionData)
...
+formulation. One λ variable is created per breakpoint (n points → n variables):
+
+    P = Σᵢ λᵢ · Pᵢ,   C = Σᵢ λᵢ · Cᵢ,   Σᵢ λᵢ = on_status,   λᵢ ∈ [0, 1]
+
+For convex cost curves this LP relaxation is tight with no extra constraints. For
+non-convex curves, an SOS2 constraint is added to enforce that at most two neighboring
+λ values are nonzero — which introduces binary variables and a MILP. If your curve is
```
GabrielKS: yeah this is extremely implementation dependent and doesn't seem to relate at all to what a `PiecewiseLinearData` actually _is_, which is a representation of any univariate piecewise linear function.

src/function_data/function_data.jl
```diff
@@ -217,20 +256,33 @@ function Base.show(io::IO, ::MIME"text/plain", fd::PiecewiseLinearData)
...
+
+**Optimization formulation:** this data shape maps to the **delta (block-offer)**
+formulation. One δ variable is created per segment (n-1 segments → n-1 variables):
+
+    P = Σₖ δₖ + offset,   C = Σₖ δₖ · slopeₖ,   0 ≤ δₖ ≤ Pₖ₊₁ - Pₖ
+
+The per-segment upper bounds enforce ordering without SOS2, so non-convex (decreasing
+slope) curves do **not** require binary variables — the LP relaxation is always valid
+for pricing. This is the standard formulation for market offers in unit commitment
+models.
```
GabrielKS: same issue as above, plus the additional issue that `PiecewiseStepData` can also be used to define an average rate curve, not just a marginal curve

src/cost_aliases.jl
```diff
@@ -86,10 +103,26 @@ Base.show(io::IO, vc::QuadraticCurve) =
...
     PiecewisePointCurve(points::Vector{Tuple{Float64, Float64}})
 
-A piecewise linear curve specified by cost values at production points.
+A piecewise linear cost curve defined by **absolute (production, total-cost) points**.
+
+Each point is `(MW, \$/h)`. The curve linearly interpolates between them. The y-values are
+**total costs**, not marginal rates. If your data instead gives marginal rates between
+breakpoints (e.g., a bid stack), use [`PiecewiseIncrementalCurve`](@ref).
+
+**Optimization formulation:** uses the **lambda (convex combination)** formulation — one
```
GabrielKS: see above

src/cost_aliases.jl
```diff
@@ -119,18 +152,34 @@ Base.show(io::IO, vc::PiecewisePointCurve) =
...
+
+A piecewise marginal-rate curve: each segment has a constant \$/MWh rate.
+
+**This is the standard format for generator bid stacks and market offers.** The y-values
+are marginal rates (slopes), not total costs. If your data gives total cost at each output
+level, use [`PiecewisePointCurve`](@ref) instead.
 
-A piecewise linear curve specified by marginal rates (slopes) between production points. May
-have nonzero initial value.
+**Optimization formulation:** uses the **delta (block-offer)** formulation — one δ variable
```
GabrielKS: see above

src/time_series_value_curve.jl
```diff
@@ -0,0 +1,133 @@
+"""
+    TimeSeriesInputOutputCurve{T <: TimeSeriesFunctionData} <: ValueCurve{T}
+
+A time-series-backed input-output curve, directly relating the production quantity to the
+cost: `y = f(x)`. Mirrors [`InputOutputCurve`](@ref) but the function data comes from a
+time series referenced by a [`TimeSeriesKey`](@ref).
+"""
+@kwdef struct TimeSeriesInputOutputCurve{
```
GabrielKS, line 8: Similar comment to my https://github.com/NREL-Sienna/InfrastructureSystems.jl/pull/558#pullrequestreview-3956858279: why not do
```julia
@kwdef struct InputOutputCurve{Union{T, TimeSeriesFunctionData{T}} where T <:
    LinearFunctionData, QuadraticFunctionData, PiecewiseLinearData}
    ...
end
```

and same for all the other concrete `ValueCurve` types. Then you still only need three `ValueCurve` types and you could still define helpers like
```julia
is_time_varying(::ValueCurve{TimeSeriesFunctionData}) = true
is_time_varying(::ValueCurve) = false
```
and aliases like
```julia
const TimeSeriesLinearCurve =
    InputOutputCurve{TimeSeriesFunctionData{LinearFunctionData}}
```

src/cost_aliases.jl
```diff
@@ -197,3 +248,151 @@ Base.show(io::IO, vc::PiecewiseAverageCurve) =
...
         Base.show_default(io, vc)
     end
+
+# ── Time-series cost aliases ──────────────────────────────────────────────────
+# Helper: format a TimeSeriesKey or Nothing for compact show output.
+_ts_key_repr(key::TimeSeriesKey) = repr(get_name(key))
+_ts_key_repr(::Nothing) = "nothing"
+
+"""
+    TimeSeriesLinearCurve
```
GabrielKS, line 231: See another comment regarding a different design suggestion for `ValueCurve`

GabrielKS (commented):
First pass through this: two high-level things:

1. The documentation is now written as if the only use case is PowerSystems -> today's PowerSimulations formulations. I don't think we want to say stuff about optimization problem formulations here, and I also don't really think we want to say that much stuff about _power systems_ stuff here, though you may disagree with the second part.
2. It seems like the time series-ification of `ValueCurve` could be done in a way that does not introduce more structs, and that this would be cleaner. I am less confident about that here than I am on #558 but it does seem doable.

src/function_data/function_data.jl
jd-lara: These distinctions need to be made in some way, if you want in a softer language but there has to be a connection between the data representation and where does it come from and why we support this.

luke-kiernan:
I agree with Gabriel's comments on the docstrings. The ones that seem definitely worth addressing:

- [increasing inaccuracy](https://github.com/NREL-Sienna/InfrastructureSystems.jl/pull/559#discussion_r2943028299)
- `PiecewiseStepData` can be for marginal or average rate curves [here](https://github.com/NREL-Sienna/InfrastructureSystems.jl/pull/559#discussion_r2943054124)
- design decisions

The rest could wait...but I'd vote for dealing with them. (Seems easy enough to set an LLM to the task: could do so myself, if desired.)

src/production_variable_cost_curve.jl
jd-lara: I agree with the implementation of this

src/production_variable_cost_curve.jl
```diff
@@ -17,9 +17,30 @@ get_function_data(cost::ProductionVariableCostCurve) =
...
+    throw(
+        ArgumentError(
+            "Concavity is not defined for time-series-backed ValueCurve; use time-series specific analysis instead.",
+        ),
+    )
+end
 is_concave(cost::ProductionVariableCostCurve) = is_concave(get_value_curve(cost))
+"Check if the cost curve is backed by time series data"
+is_time_series_backed(cost::ProductionVariableCostCurve) =
+    is_time_series_backed(get_value_curve(cost))
```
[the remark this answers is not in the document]
jd-lara: Bad suggestion

luke-kiernan (commented):
So you decided to go with the parametric type approach after all?

jd-lara:
> So you decided to go with the parametric type approach after all?

For one of the types made sense, @GabrielKS was right for the other with the nested types it didn't

luke-kiernan:
Decided to merge it: we've started to use it in IOM, so it'd be confusing to not do so.

PR #561, opened 2026-04-01
Prohibit the use of / in component names

daniel-thom:


jd-lara:
@daniel-thom to avoid breaking users during the current version I suggest we do this in IS4.

src/components.jl
```diff
@@ -1,5 +1,35 @@
 const ComponentsByType = Dict{DataType, Dict{String, <:InfrastructureSystemsComponent}}
 
+# '/' is the HDF5 path separator and cannot appear in dataset names.
+# Refer to https://github.com/NREL-Sienna/PowerSystems.jl/issues/1647 for the trigger of
+# this rule. The '/' character could also cause problems if we ever wanted to use the
+# component name as a file or directory name.
+const INVALID_COMPONENT_NAME_CHARACTERS = ('/',)
```
jd-lara, line 7: @josephmckinsey we need to enforce this somewhere in the DB and Schema for the existing versions that's fine but in the future IS4 and PSY6 we should eliminate it

jd-lara:
@daniel-thom we can close this and handle it in IS4 later since it is a corner case

PR #563, opened 2026-04-08
add interval to ts interfaces

jd-lara:


src/time_series_metadata_store.jl
```diff
@@ -653,25 +657,44 @@ end
...
+            horizon
+            ,initial_timestamp
+            ,interval
+            ,resolution
+            ,window_count
+        FROM $ASSOCIATIONS_TABLE_NAME
+        WHERE $where_clause
+        LIMIT 1
+    """
+    table = Tables.rowtable(_execute(store, query, params))
```
daniel-thom: Can we use `_execute_cached`? I do not think this is called in a performance sensitive area, but it was there before.

jd-lara: I just pushed the changes

daniel-thom (approved):

PR #565, opened 2026-04-14
namedtuple time series

jd-lara:
This feature will allow to track named tuples as time series more cleanly that the existing implementation in PSY


src/tuple_time_series.jl
```diff
@@ -0,0 +1,143 @@
...
+the raw tuple with the field names carried by `T`.
+"""
+function build_static_tuple(
+    tts::TupleTimeSeries{T},
+    owner::TimeSeriesOwners,
+    start_time::Dates.DateTime,
+) where {T <: NamedTuple}
+    key = get_time_series_key(tts)
+    vals = get_time_series_values(owner, key; start_time = start_time, len = 1)
+    raw = vals[1]
```
luke-kiernan: Nitpick: use `only`.
```suggest
    raw = only(get_time_series_values(owner, key; start_time = start_time, len = 1))
```

src/tuple_time_series.jl
```diff
@@ -0,0 +1,143 @@
...
+# ── Serialization ────────────────────────────────────────────────────────────
+
+function add_serialization_metadata!(
+    data::Dict,
+    ::Type{TupleTimeSeries{T}},
+) where {T <: NamedTuple}
+    data[METADATA_KEY] = Dict{String, Any}(
+        TYPE_KEY => "TupleTimeSeries",
+        MODULE_KEY => string(parentmodule(TupleTimeSeries)),
+        "namedtuple_fields" => [string(n) for n in fieldnames(T)],
```
luke-kiernan: Nitpick: you could store the value associated with `"namedtuple_fields"` as a tuple, instead of as a vector. Might compile better.

src/tuple_time_series.jl
```diff
@@ -0,0 +1,143 @@
...
+# supporting as first-class shapes; PSY precompiles against these to keep the hot
+# construction path branchless. The generic fallback below handles any other arity
+# by validating at runtime.
+_validate_tuple_time_series_type(
+    ::Type{T},
+) where {Names, T <: NamedTuple{Names, NTuple{2, Float64}}} = nothing
+
+_validate_tuple_time_series_type(
+    ::Type{T},
+) where {Names, T <: NamedTuple{Names, NTuple{3, Float64}}} = nothing
```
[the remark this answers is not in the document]
luke-kiernan: Is there a way to work around this (CoPilot's concern) with `@generated`? Still mildly messy, though: the `@generated` calls (that create the no-validation constructors for your specific named tuples you want to support) would happen in PSY, so they probably couldn't be inlined into IS.

src/tuple_time_series.jl
```diff
@@ -0,0 +1,143 @@
...
+) where {T <: NamedTuple}
+    data[METADATA_KEY] = Dict{String, Any}(
+        TYPE_KEY => "TupleTimeSeries",
+        MODULE_KEY => string(parentmodule(TupleTimeSeries)),
+        "namedtuple_fields" => [string(n) for n in fieldnames(T)],
+    )
+    return
+end
+
+function deserialize(::Type{TupleTimeSeries}, data::Dict)
```
luke-kiernan, line 135: Do you know what `TupleTimeSeries` you're deserializing into, or is that something you have to recover from the data itself? This is type unstable:
```Julia
N = length(field_names) # runtime value
T = NamedTuple{field_names, NTuple{N, Float64}} # not inferrable
# ...
return TupleTimeSeries{T}(key) # not inferrable
```
But if you're recovering the type from the data, then there's really nothing you can do.

luke-kiernan (changes requested):
A couple compilation/type stability comments.

src/tuple_time_series.jl
```diff
@@ -0,0 +1,143 @@
...
+`N`; the inner constructor rejects other shapes.
+
+# Example
+```julia
+const StartUpStages = NamedTuple{(:hot, :warm, :cold), NTuple{3, Float64}}
+tts = TupleTimeSeries{StartUpStages}(ts_key)
+stages = build_static_tuple(tts, component, start_time)  # returns a StartUpStages
+```
+"""
+struct TupleTimeSeries{T <: NamedTuple} <: InfrastructureSystemsType
```
daniel-thom, line 24: @jd-lara About 1 (2?) year ago you, Gabriel, and I discussed an idea: parameterize the time series type by the type of its values, e.g. `SingleTimeSeries{Float64}` (analogous to `Vector{Float64}`). Is now the time to do it? It would solve this problem in a better way (the time series key has the type).

luke-kiernan (approved):
A couple small issues to be addressed in follow-up PRs--will open issues for those.

luke-kiernan (approved):

src/tuple_time_series.jl
jd-lara: What would it take to do this? What's a reasonable time line to tackle it? We can include it in IS4

PR #570, opened 2026-04-15
Lk/units prototype rebase

luke-kiernan:
Add unitful setters and getters. The new dependency (which I named [PowerSystemsUnits.jl](https://github.com/luke-kiernan/PowerSystemsUnits.jl), open to other suggestions) could be folded into IS, but I'd prefer to keep it separate.


jd-lara (commented):
Move the code to be a Module in IS or split the functionality between IS and PSY accordingly.

luke-kiernan:
Superseded by #577

PR #574, opened 2026-04-17
Lk/units domain agnostic

luke-kiernan:
Backend for explicit units in getters and setters.


jd-lara:
One comment I see is that we need The five TimeSeries*Curve aliases to full parametric type. Please add overrides for symmetry and better printing.

src/utils/generate_structs.jl
```diff
@@ -160,12 +189,18 @@ function generate_structs(directory, data::Vector; print_results = true)
...
+                        "conversion_unit" => conversion_unit,
                     ),
                 )
             end
             if field["name"] != "internal" && accessor_module == ""
-                push!(unique_accessor_functions, accessor_name)
+                if include_getter
+                    push!(unique_accessor_functions, accessor_name)
+                end
+                # Always export setter name even if exclude_setter is true,
```
jd-lara: exclude_setter still registers the public name, but exclude_getter does not.

luke-kiernan: Yeah it should be exported in both cases. Will fix.

luke-kiernan:
Closing because it's superseded by #577

PR #575, opened 2026-04-20
is_time_series_backed on `FuelCurve`

luke-kiernan:
There are some test failures, but they're unrelated: it's module prefix vs no module prefix discrepancies in the `sprint` and `repr` tests


jd-lara (approved):
Seems like a reasonable approach

PR #586, opened 2026-06-18
add instructions about `TestEnv.jl` to `Sienna.md`

luke-kiernan:


luke-kiernan:
Turns out `using TestEnv; TestEnv.activate(); push!(ARGS, "test_file_name"); include("test/runtests.jl")` doesn't work. Will revise

jd-lara (approved):

jd-lara:
> Turns out `using TestEnv; TestEnv.activate(); push!(ARGS, "test_file_name"); include("test/runtests.jl")` doesn't work. Will revise

I think we need to call the tests files with a different strategy

jd-lara (approved):

PR #594, opened 2026-07-06
Speed up transform_single_time_series! for many SingleTimeSeries

daniel-thom:
## Problem

`transform_single_time_series!` was very slow with tens of thousands of `SingleTimeSeries` instances. It issued roughly **4 SQLite queries per series**:

- **Check phase** (`_check_transform_single_time_series`), per series:
  1. `get_forecast_parameters` — even though it depends only on `(resolution, interval)`, which is nearly constant across series.
  2. `list_metadata` for an existing `Deterministic` forecast — run on every component, including those with no forecasts.
  3. `list_metadata` for an existing `DeterministicSingleTimeSeries` (when `skip_existing`).
- **Add phase**: one INSERT per series via the single-row `add_metadata!`.

At 50k series that's ~200k queries.

## Changes

**Check phase**
- Memoize `get_forecast_parameters` by `(resolution, interval)`.
- Fetch once the set of owner UUIDs that actually own a `Deterministic` (and, when `skip_existing`, a `DeterministicSingleTimeSeries`) forecast, and only run the per-component metadata query for owners in those sets. When no such forecasts exist, all per-component queries are skipped. This is exact — an owner not in the set could not have matched the old query.

**Add phase**
- Add a vectorized `add_metadata!(store, owners, all_metadata)` that issues a single `executemany` INSERT instead of one INSERT per series, and use it from `_transform_single_time_series!`. `executemany` uses a savepoint when already inside `begin_time_series_update`'s transaction, so nesting is safe.
- Factor the associations column list into `ASSOCIATIONS_TABLE_COLUMNS`.

Behavior is preserved: conflict detection and `skip_existing` idempotency are unchanged; only provably non-matching queries are avoided.

## Verification

- **Benchmark** (10,000 `SingleTimeSeries`, post-warmup): **~3.67s → ~0.40s (~9×)**, with better scaling as the count grows (old per-component cost was linear).
- Full test suite passes (8372 tests, zero error-level log events). Formatter run.

### DIFF
--- a/src/system_data.jl
+++ b/src/system_data.jl
@@ -683,11 +683,9 @@ function _transform_single_time_series!(
 
     try
         begin_time_series_update(data.time_series_manager) do
-            add_metadata!(
-                data.time_series_manager.metadata_store,
-                components,
-                all_metadata,
-            )
+            for (component, metadata) in zip(components, all_metadata)
+                add_metadata!(data.time_series_manager.metadata_store, component, metadata)
+            end
         end
     catch
         # Only remove the metadata entries that were added in this batch so that
@@ -708,6 +706,12 @@ function _transform_single_time_series!(
     end
     return
 end
+
+# Replicates the SQL partial feature match used by list_metadata: every requested
+# feature key-value pair must be present in the existing metadata's features.
+_features_contain(existing_features, requested_features) =
+    all(kv -> isequal(get(existing_features, kv.first, nothing), kv.second),
+        requested_features)
 
 """
 Check that all existing SingleTimeSeries can be converted to DeterministicSingleTimeSeries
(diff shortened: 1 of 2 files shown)


src/time_series_metadata_store.jl
```diff
@@ -520,6 +522,57 @@ function add_metadata!(
...
+duplicates.
+
+This is much faster than calling `add_metadata!` in a loop when adding a large number of
+entries because it issues one `executemany` INSERT instead of one INSERT per entry.
+"""
+function add_metadata!(
+    store::TimeSeriesMetadataStore,
+    owners::Vector{<:TimeSeriesOwners},
+    all_metadata::Vector{<:TimeSeriesMetadata},
+)
```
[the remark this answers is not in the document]
daniel-thom: Addressed in 9443d04e4 — widened both parameters to `AbstractVector{<:...}`.

src/time_series_metadata_store.jl
```diff
@@ -520,6 +522,57 @@ function add_metadata!(
...
+            )
+        end
+        sfm = get_scaling_factor_multiplier(metadata)
+        rows[i] = _create_row(
+            metadata,
+            owner,
+            _get_owner_category(owner),
+            _convert_ts_type_to_string(time_series_metadata_to_data(metadata)),
+            make_features_string(metadata.features),
+            isnothing(sfm) ? missing : JSON3.write(serialize(sfm)),
```
josephmckinsey: Should we extract out this logic since it's duplicated between both add_metadata! functions?

src/system_data.jl
```diff
@@ -755,34 +794,38 @@ function _check_transform_single_time_series(
...
         # DeterministicSingleTimeSeries with the same name, resolution, features,
         # horizon, and interval.
-        if skip_existing
+        if skip_existing && item.owner_uuid in owners_with_deterministic_sts
+            ts_features_symbols = Dict{Symbol, Any}(
+                Symbol(k) => v for (k, v) in get_features(item.metadata)
+            )
             existing = list_metadata(
-                data.time_series_manager.metadata_store,
+                store,
```
josephmckinsey: Since this still keeps a potential O(n) number of SQL queries, this could still bite us badly if someone actually calls transform_single_time_series! twice with skip_existing=true, since it would hang just as badly. Should that just be user error or would a warning be appropriate?

An alternative implementation would be to write a query joining all single time series to all "compatible" deterministic/deterministic single time series instead of searching each time, or creating a dict to search instead. I imagine that we will be replacing this once the new time series library is available, in which case, maybe just a warning?

josephmckinsey (commented):
Looks good enough as is, but there is some minor code duplication and potential future problems.

jd-lara:
@daniel-thom what's the status of this?

src/time_series_metadata_store.jl
daniel-thom: This is now obsolete. The bulk `add_metadata!` method added by Claude in round 1 was unnecessary. The SQLite.jl implementation of `executemany` calls `execute` in a for loop, so there was no point.

src/system_data.jl
daniel-thom: Good catch — that was a real hole in the first version. The owner-UUID pre-filter only helped when few owners had forecasts, and after one `transform_single_time_series!` every owner is in the `DeterministicSingleTimeSeries` set, so a second call fell straight back to N per-component queries. The idempotent re-run was the worst case, not the best one.

I took the dict approach you suggested rather than a warning. `_check_transform_single_time_series` now fetches all `Deterministic`/`DeterministicSingleTimeSeries` metadata in a single `list_metadata_with_owner_uuid` call and indexes it by `(owner_uuid, name, resolution)`; both the conflict check and the `skip_existing` check run in memory against that lookup. The per-iteration `list_metadata` calls are gone.

There is now no SQL per series at all in that loop: `get_forecast_parameters` is memoized on `(resolution, interval)`, `get_component` is an in-memory `component_uuids` lookup, and `_check_single_time_series_transformed_parameters` is pure metadata arithmetic. The check phase is two bulk queries plus one `get_forecast_parameters` per distinct `(resolution, interval)`, regardless of series count — so a second call costs the same as the first and a warning doesn't seem warranted. `check_transform_single_time_series` gets the same benefit, since it runs the check with `skip_existing = true`. (For anyone reading this later: the public spelling of this path is `delete_existing = false`.)

Two trade-offs worth naming. Memory: the check now materializes all forecast metadata up front, alongside the `SingleTimeSeries` list it already held, so peak metadata during a re-run roughly doubles. Semantics: the `skip_existing` predicate is now an exact feature match in memory (`_features_contain`) instead of SQL `LIKE`, which is stricter — it can skip fewer entries but never more. Copilot raised that same point on line 714 and I'm treating it as a separate thread.

And agreed that this is a stopgap until the new time-series library lands.

jd-lara (approved):

PR #597, opened 2026-07-22
Feat/associations in rust store

daniel-thom:


src/supplemental_attribute_associations.jl
```diff
@@ -1,794 +1,351 @@
...
-    # If you add any fields, ensure they are managed in deepcopy_internal below.
-end
+# Consequences worth knowing:
+#   - Associations are persisted by the store's `<path>.nc` / `<path>.sqlite` pair, so a
+#     system with supplemental attributes but no time series now produces a storage
+#     artifact where it previously produced none. `isempty(::RustTimeSeriesStore)`
+#     accounts for association rows so `serialize` writes that artifact.
+#   - The store enforces uniqueness on the `(component_id, attribute_id)` pair. The old
+#     table had no constraint and relied on callers checking first; callers still check,
+#     so the error message can name the objects, but a missed check is now caught.
```
daniel-thom, line 14: @jd-lara I want to give you a chance to comment here. I want to consolidate the supplemental attribute association logic in the Rust store. The logic is currently duplicated in the Python and Julia libraries. Moving to Rust is natural because that is where all SQLite code is located. It means that IS and infrasys can remove dependencies on SQLite.

I'm contemplating a name change to the Rust store to go along. Instead of only storing time series, it becomes a "catalog and store" -> castore. Package names (rust/julia/python): castore, castore, Castore.jl

I think it is an overwhelmingly positive addition in scope. The only downside is on lines 8-9 above.

daniel-thom: There is a middle-ground compromise. Suppose we do keep the SQLite in Rust. On serialization, IS.jl could remove all records from SQLite and store them in the JSON file, preserving the current behavior. It might seem odd in the code, but would be fine.

jd-lara: This is not ideal for the cases where we have systems with attributes and not time series. Is there any strong reason to purpuse the increase in scope at this time?

daniel-thom: Reasons:
- Drop the dependency on SQLite.jl. Use the much better supported SQLite library in Rust.
- Consolidate logic from the Python and Julia libraries in one place.

As I mentioned in my second comment, we can preserve the current behavior by calling the remove function during serialization to remove all records from the SQLite file. From the perspective of the Rust store, we are removing/adding during serde.

daniel-thom: I am merging into the feature branch. How to handle this issue is still open.

PR #600, opened 2026-07-27
 Replace the HDF5 + SQLite time-series backend with InfraStore (Rust core)

daniel-thom:
## Summary

This PR removes the legacy time-series storage stack — HDF5 arrays, SQLite
metadata store, and the SQLite supplemental-attribute associations adapter —
and replaces it with [InfraStore](../../infrastore), a Rust-core store
(NetCDF arrays + SQLite catalog behind an FFI) with a typed Julia surface.
Time-series data, catalog queries, forecast window slicing, and
supplemental-attribute associations all now live in the Rust store.

At 100k series, bulk adds and reads are faster than IS4,
and both stay flat with store size (IS4 degrades with N on both paths). See
`benchmark/README.md` for methodology and full results.

## Backend swap

- **Removed:** `hdf5_time_series_storage.jl`, `in_memory_time_series_storage.jl`,
  `time_series_metadata_store.jl` (~1,800 lines of SQL), `time_series_storage.jl`,
  `utils/sqlite.jl`, and the SQLite associations adapter.
- **Added:** `infrastore.jl` (store adapter), `store.jl`, `time_series_context.jl`
  (batching), `time_series_normalization.jl`.
- **Dependencies:** dropped CSV, HDF5, H5Zblosc, SQLite, Serialization, and
  Tables; added InfraStore (declared via `[sources]` by relative path until it
  is registered).
- Deduplication is now by content hash rather than UUID; shared arrays are
  reference-counted by the store.

## Breaking API changes

- **Metadata structs are gone.** `SingleTimeSeriesMetadata`,
  `DeterministicMetadata`, `ProbabilisticMetadata`, and `ScenariosMetadata`
  are collapsed into `TimeSeriesKey`; the metadata-oriented APIs are renamed
  to key-oriented ones. The per-series UUID is removed.
- **One batching primitive:** `time_series_transaction(data) do context ... end`.
  Pass `context` to `add_time_series!` calls inside the block and they are
  buffered and committed as one bulk write (block-sized array writes,
  feature-set dedup); the block is also a store transaction, so a throw rolls
  everything back, including removals. Removed: `begin_time_series_update`,
  `open_time_series_store!` (and its unused `mode` argument),
  `bulk_add_time_series!`, `TimeSeriesAssociation`, and
  `ADD_TIME_SERIES_BATCH_SIZE`.
- **Type hierarchy is parameterized on the value element type:**
  `TimeSeriesData{T}`, with concrete structs carrying `{T, N}` (array rank).
  Payloads are plain `Array`s instead of `TimeArray`s. `TupleTimeSeries` is
  removed in favor of tuple-valued element types on the standard structs.
- **Components and supplemental attributes are identifie
  instead of UUIDs; the `uuid` field is removed from
  `InfrastructureSystemsInternal`.
- **CSV time-series parsing is removed** (`time_series_formats.jl`,
  `time_series_parser.jl`, and the associated test data)
- In-memory stores serialize via `persist!`; the reconstruction tag is stored
  in `ext` as JSON.

## New capabilities

- Full support for `NonSequentialTimeSeries`.
- `ForecastReader` for shared-data forecast iteration and a static
  time-series reader.
- `get_time_series_hash` / `get_shared_time_series`.
- Forecast window slicing and catalog queries pushed int
  (server-side `time_range` slicing; no full-array materialization).
- `DeterministicSingleTimeSeries` reads materialize as `

## Setup / requirements

- Requires `INFRASTORE_LIB` pointing at a built `libinfr
- InfraStore is not yet registered; the `[sources]` entry resolves it by
  relative path (`../../infrastore/julia/InfraStore.jl`)
  packages that dev IS by path resolve it automatically. Drop once registered.

## Testing

- Full suite: 8,656 passing, 1 broken (2026-07-26 baseline).
- Benchmarks vs IS4 included under `benchmark/` with res

## Downstream impact

Every psy6 package sits on IS; this is the highest-risk
(time-series + serialization). PowerSystems, PowerSystemCaseBuilder, and the
other consumers need coordinated updates for the removed
the batching-primitive rename, integer ids, and the InfraStore source entry.


benchmark/results_is4.csv
```diff
@@ -0,0 +1,8 @@
+branch,scenario,op,n,total_s,us_per_op,bytes
```
daniel-thom, line 1: This will be deleted before merging.

PR #602, opened 2026-08-06
Include concrete type in hash_from_fields

luke-kiernan:
## Problem

`hash_from_fields` hashed only an instance's fields, never its type. Since IS4 moved `power_units` from a runtime field to the type parameter `U <: AbstractUnitSystem`, `CostCurve{T, NaturalUnit}` and `CostCurve{T, SystemBaseUnit}` with the same value curve have identical fields, so they hashed identically. A bit of an edge case, but potentially bad for hash-based collections.

The same held for any two types with matching field names/values, e.g. `IncrementalCurve` vs `AverageRateCurve`.

## Fix

Hash the concrete type alongside the fields, and take it as a method type parameter (`hash_from_fields(a::T, h::UInt) where {T}`) instead of calling `typeof` inline.

This covers all three `Base.hash` methods routed through it: `ProductionVariableCostCurve`, `ValueCurve`, and `FunctionData`.

## Testing

New testset `hash distinguishes unit systems and curve types` in `test/test_cost_functions.jl`
— 5 assertions, passing. Full suite not run (change is small and localized); CI will cover it.

### DIFF
--- a/src/utils/utils.jl
+++ b/src/utils/utils.jl
@@ -761,11 +761,17 @@ double_equals_from_fields(a::T, b::T) where {T} =
 isequal_from_fields(a::T, b::T) where {T} =
     compare_over_fields(isequal, &, true, a, b)
 
-"Compute a hash of the instance `a` by combining hashes of all its fields"
+"""
+Compute a hash of the instance `a` by combining hashes of all its fields along with its
+concrete type. The type must be included because instances that differ only in type
+parameters carrying no field data -- e.g. the `U <: AbstractUnitSystem` marker of
+`CostCurve{T, U}` -- would otherwise hash identically while comparing unequal.
+"""
 hash_from_fields(a) = hash_from_fields(a, zero(UInt))
 
-function hash_from_fields(a, h::UInt)
-    for field in sort!(collect(fieldnames(typeof(a))))
+function hash_from_fields(a::T, h::UInt) where {T}
+    h = hash(T, h)
+    for field in sort!(collect(fieldnames(T)))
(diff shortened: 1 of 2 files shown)


luke-kiernan:
I replicated the test failure locally on IS4, so it's unrelated to these changes

src/utils/utils.jl
```diff
@@ -761,11 +761,17 @@ double_equals_from_fields(a::T, b::T) where {T} =
...
+Compute a hash of the instance `a` by combining hashes of all its fields along with its
+concrete type. The type must be included because instances that differ only in type
+parameters carrying no field data -- e.g. the `U <: AbstractUnitSystem` marker of
+`CostCurve{T, U}` -- would otherwise hash identically while comparing unequal.
+"""
 hash_from_fields(a) = hash_from_fields(a, zero(UInt))
 
-function hash_from_fields(a, h::UInt)
-    for field in sort!(collect(fieldnames(typeof(a))))
+function hash_from_fields(a::T, h::UInt) where {T}
```
daniel-thom, line 772: Does it need to be recursive?

luke-kiernan: > Does it need to be recursive?

You'd prefer that I write a bespoke non-recursive hash method for the handful of cases where the type parameters matter? This function was already recursive to begin with--for nested structs, the `Base.hash` of the field routes back to `hash_from_fields`--and I don't see a way around that in general...

daniel-thom: Nevermind, I misread. Sorry.

daniel-thom (approved):

PR #604, opened 2026-08-07
auto generate updates

jd-lara:
This PR makes some updates about the code autogeneration that was moved to PSY. It is part of the serde updates for psy6


docs/src/docs_best_practices/how-to/compile.md
```diff
@@ -26,12 +26,15 @@ documentation.
...
+`PowerSystems.jl` has its own struct generator (`PowerSystems.StructGeneration`, forked from
+this package so PSY-specific generation stays out of IS) — use it, not
+`InfrastructureSystems.generate_structs`, for PSY's descriptor. From a terminal at the root of
+the `PowerSystems.jl` repository, run:
 
 ```
 julia --project=.
-using InfrastructureSystems
-InfrastructureSystems.generate_structs(
+using PowerSystems
```
daniel-thom, line 36: If this section is included at all, it should point to IS code, not PSY. I'll note, however, that with the migration to infrastore, there will be no auto-generated structs in IS, so you could just delete this section.

jd-lara:
@daniel-thom then we can close this PR if the code autogeneration is going away with infra-store

PR #608, opened 2026-08-17
Move time series management to InfraStore.jl

daniel-thom:


benchmark/bench_branch.jl
jd-lara, line 1: Do we need to check in all the benchmarking?

src/time_series_manager.jl
```diff
@@ -25,189 +14,223 @@ function TimeSeriesManager(;
...
+        auto_flush_bytes = auto_flush_bytes,
+    )
+    begin_transaction!(context)
+    result = try
+        func(context)
+    catch
+        discard!(context)
+        rethrow()
     end
+    commit!(context)
```
jd-lara: I don't understand the commit! outside the try/catch. Commit! flushes and commits with only a finally context.closed = true. So when the block succeeds but the final flush throws, discard! never runs and the InfraStore transaction is left open holding the store's write lock.

src/time_series_interface.jl
```diff
@@ -967,98 +847,47 @@ See also: [`get_time_series_array`](@ref get_time_series_array(
...
 """
 function has_time_series(owner::TimeSeriesOwners; kwargs...)
     mgr = get_time_series_manager(owner)
     isnothing(mgr) && return false
-    return has_metadata(mgr.metadata_store, owner; kwargs...)
+    kw = Dict(kwargs)
+    name = pop!(kw, :name, nothing)
+    T = pop!(kw, :time_series_type, TimeSeriesData)
+    isnothing(name) && return infrastore_has_any(owner; time_series_type = T)
+    return infrastore_has_time_series(
```
jd-lara: Why is this still here if `infrastore_has_time_series` already handles an abstract T correctly

jd-lara (commented):
The PR is large but these are issues that concern me more

src/system_data.jl
```diff
@@ -97,88 +108,58 @@ function SystemData(
...
-        func,
-        get_data_store(get_time_series_manager(data)),
-        mode,
-        args...;
-        kwargs...,
-    )
+"""
+Return the next integer id to assign to a component and advance the component counter.
+"""
+function get_next_component_id!(data::SystemData)
```
daniel-thom: We have an argument from @jd-lara to store one stream of IDs across components and supplemental attributes.

benchmark/bench_branch.jl
daniel-thom: I can squash them out.

src/time_series_interface.jl
```diff
@@ -384,8 +350,6 @@ function get_time_series_array(
         resolution = get_resolution(key),
```
jd-lara: These key based paths forward `resolution` and the features but never `interval`, so a `ForecastKey` that differs only by interval reads and removes the wrong series. Can we forward `interval = get_interval(key)` here and in `get_time_series_timestamps`, `get_time_series_values`, `_copy_time_series!` and `remove_time_series!`?

jd-lara:
`to_json` and `from_json` are going away in favor of the OpenAPI `to_openapi`/`from_openapi` converters, so we should not add callers or fix anything in that path here. The pair in `utils/recorder_events.jl` is unrelated, it reads and writes the line delimited event log, so those stay, but we should rename them to something like `write_recorder_to_json` and `read_recorder_from_json` so they are not confused with the ones we are removing.

jd-lara (approved):

PR #609, opened 2026-08-18
Spell per-unit tags out for display; per-column table units

luke-kiernan:
## What changed

Two display-side changes that [PowerSystems.jl#1753](https://github.com/Sienna-Platform/PowerSystems.jl/issues/1753) needs. Paired with the PSY PR that consumes them.

- **`display_string(x)`** renders a value with its relative-unit tag spelled out instead of the terse `DU`/`SU`. Those are convenient shorthand but aren't standard terminology, so verbose output (a component's `text/plain` display) spells them out while terse contexts — the one-line `show`, table cells — keep the tags.

  ```
  0.6 * DU                          → "0.6 p.u. in device base"
  (min = 0.0 * SU, max = 2.5 * SU)  → "(min = 0.0 p.u., max = 2.5 p.u.) in system base"
  ```

  A compound field whose elements share one base states it once, after the tuple. Mixed or partially-tagged tuples have nothing to hoist, so they fall back to spelling each element out.

- **`show_components`' `units` kwarg** now accepts a column-to-unit mapping (`AbstractDict` or `NamedTuple`) as well as a single unit, so one table can show one column in natural units and another in device base. A column absent from the mapping keeps its own `display_units_arg` default rather than inheriting a neighbour's.

## How

`display_string` splits the old per-marker string into a value part and a base label, which is what lets the base be factored out of a `NamedTuple`. Column resolution moves the old `units === nothing ? trait_arg : units` ternary into a small `_column_units` dispatch, so the mapping forms are three methods rather than a branch.

### DIFF
--- a/src/internal.jl
+++ b/src/internal.jl
@@ -8,7 +8,7 @@ Unit system for component data values.
 
 # Values
 - `SYSTEM_BASE`: Per-unit values on the system base power
-- `DEVICE_BASE`: Per-unit values on the device base power
+- `DEVICE_BASE`: Per-unit values on the component base power
 - `NATURAL_UNITS`: Values in natural units (e.g., MW, MVAR)
 """ UnitSystem
 
--- a/src/relative_units.jl
+++ b/src/relative_units.jl
@@ -1,7 +1,7 @@
 ###############################
 # Relative (per-unit) markers and RelativeQuantity wrapper.
 #
-# These types are domain-agnostic — they express "device base" / "system base"
+# These types are domain-agnostic — they express "component base" / "system base"
 # / "natural unit" without assuming any particular physical domain. Downstream
 # packages (e.g. PowerSystems) attach domain-specific meaning via categories
 # and conversions.
@@ -64,7 +64,7 @@ A quantity tagged with a per-unit marker.
 
 # Examples
 ```julia
-0.6 * DU  # 0.6 per-unit on device base
+0.6 * DU  # 0.6 per-unit on component base
 0.3 * SU  # 0.3 per-unit on system base
 ```
 """
@@ -245,7 +245,7 @@ Base.show(io::IO, ::NaturalUnit) = print(io, "NU")
     display_string(x) -> String
 
 Render `x` for human-facing display, spelling relative-unit tags out in full
-("0.6 p.u. in device base") where `show` prints the terse "0.6 DU". `DU`/`SU`
+("0.6 p.u. in component base") where `show` prints the terse "0.6 DU". `DU`/`SU`
 are convenient to type but are not standard terminology, so verbose output
 (e.g. a component's `text/plain` display) spells them out; terse contexts such
 as tabular cells keep the short tags.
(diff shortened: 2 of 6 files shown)


jd-lara:
this work needs to change the semantics to component based from device based as choosen by the Sienna team

luke-kiernan:
> this work needs to change the semantics to component based from device based as choosen by the Sienna team

Okay so...I should revise `p.u. in device base` to `p.u. in component base`? Are we planning to similarly change the enum `DEVICE_BASE` (`COMPONENT_BASE`) and the shorthand `DU` (`CU`) too?

jd-lara:
Yes, the default semantics are now component base as requested by the NLR. - Sienna group

luke-kiernan:
ok I will change the pretty printing semantics here

luke-kiernan:
imo this is ready to merge. I changed `device base` to `component base` in the pretty printing but left the enum and type renames for a separate PR (bc that'll break stuff downstream)

src/internal.jl
```diff
@@ -8,7 +8,7 @@ Unit system for component data values.
 
 # Values
 - `SYSTEM_BASE`: Per-unit values on the system base power
-- `DEVICE_BASE`: Per-unit values on the device base power
+- `DEVICE_BASE`: Per-unit values on the component base power
```
daniel-thom, line 11: Should this be COMPONENT_BASE?

src/relative_units.jl
```diff
@@ -20,6 +20,7 @@ export RelativeQuantity
 export DU, SU, NU
```
daniel-thom, line 20: CU?

src/relative_units.jl
```diff
@@ -20,6 +20,7 @@ export RelativeQuantity
...
 export display_units_arg
```
daniel-thom, line 21: We have typically not exported names from IS. Is there a reason to start doing it?

test/test_printing.jl
```diff
@@ -86,6 +86,34 @@ end
     @test !isnothing(val2_pos) && !isnothing(val_pos)
     @test first(val2_pos) < first(val_pos)
 
+    # A column-to-unit mapping sets units per column; a column missing from the
+    # mapping keeps its own trait default rather than inheriting a neighbour's.
```
daniel-thom:
```suggestion
    # mapping keeps its own trait default rather than inheriting a neighbor's.
```
Do we need to tell Claude that we are in the US?

src/relative_units.jl
luke-kiernan: Same as above--I can do the rename, I just don't want to do it here

src/internal.jl
luke-kiernan: Deferred to a separate PR. Easier to keep "this actually adds new functionality" separate from "this is just a semantic change." Also DU -> CU is breaking which means higher level of inter-repo coordination on the merges

src/relative_units.jl
luke-kiernan: I believe this is just exporting from the `RelativeUnits` submodule, not from IS as a whole...

jd-lara (approved):

PR #610, opened 2026-08-20
Add bulk catalog reads and OpenAPI converters for both association tables

jd-lara:
This is the companion to https://github.com/NatLabRockies/infrastore/pull/39 among other updates to support the use of the Sienna Schemas with Infrastore. 

The schemas adopted are based heavilty on the infrastore specification to avoid conflicts.


jd-lara:
@daniel-thom thet tests will fail without the release of infrastore

daniel-thom (approved):

PR #611, opened 2026-08-20
Match the full key identity in key-addressed remove, read, and hash paths

daniel-thom:
Follow-up to the key-addressed interval work on `feat/infrastore-integration` (targets that branch, part of #608's line). A verified correctness review found that three key-addressed paths still resolved a fully-resolved `TimeSeriesKey` through by-name queries whose feature matching is subset-based, plus one interval miss and a stale docstring contract.

## Fixes

- **Keyed `remove_time_series!` deleted too much.** With series A `{scenario: "a"}` and sibling B `{scenario: "a", model: "x"}` (a legal pair — the store identity hashes the full feature set), removing by A's key deleted both. It now calls a new exact-key primitive `infrastore_remove_time_series!` built on the store's typed removal, which matches the whole feature-set hash plus name/type/resolution/interval. Behavior change: a keyed removal of a series that is not stored now throws an actionable `ArgumentError` instead of silently no-opping (the old count-returning filter swallowed it).
- **Keyed `get_time_series_array` / `get_time_series_timestamps` / `get_time_series_values` threw spurious ambiguity errors.** They re-resolved the key by name (subset feature matching), so an exact key failed with "Found more than one matching metadata" whenever a superset-features sibling existed — even though `get_time_series(owner, key)` on the same key succeeded. They now read key-addressed and delegate to the instance-form accessors, which also removes the per-call `Dict{Symbol, Any}` features allocation, the kwarg splat, and the redundant catalog re-resolution (and kills the key-field drift class the interval commit had to patch at every unpack site).
- **`infrastore_get_time_series_hash` ignored the interval.** With two forecasts differing only by interval (legal as of this branch), it returned whichever catalog row iterated first. It now filters on the interval via a two-method dispatch helper, matching the bulk variant.
- **`TimeSeriesKey` docstring understated the required interface.** The keyed paths call `get_interval` and `get_features` unconditionally with no abstract fallback; the docstring now lists the real contract and states that the concrete key set is closed.

## Verification

The exact-match claim was verified against the core (read-only): `infrastore_store_remove_typed` builds a `KeyIdentity` and `delete_by_key` matches `features_hash = ?` exactly, erroring `NotFound` when nothing matches.

New tests: a superset-features-sibling fixture covering the keyed read trio (including `start_time`/`len` slicing) and removal exactness in both directions plus the not-found error, and per-interval hash assertions on the existing two-interval fixture.

Full suite: 8930 pass, 0 fail, 3 pre-existing broken markers. Formatter clean.

### DIFF
--- a/src/infrastore.jl
+++ b/src/infrastore.jl
@@ -678,8 +678,11 @@ function infrastore_remove_time_series!(
     catch e
         # `catch`-block exception inspection: the core reports both conditions
         # through its own error types, which IS maps to the errors callers
-        # dispatch on.
-        if e isa InfraStore.InvalidParameterError
+        # dispatch on. The DST-orphan guard can only fire for a SingleTimeSeries
+        # key; an InvalidParameterError for any other key type is something else
+        # and propagates as itself.
+        if e isa InfraStore.InvalidParameterError &&
+           get_time_series_type(key) <: SingleTimeSeries
             throw(
                 ArgumentError(
                     "Cannot remove SingleTimeSeries '$name' because it is attached to a " *
@@ -1757,11 +1760,12 @@ _decode_element(element, encoding::RowEncoding) =
 
 """Route `has_time_series(owner, T, name; ...)` to the InfraStore store. Honors partial
 (subset) feature / resolution queries: matches if any stored series of type `T`
-contains at least the requested features."""
+contains at least the requested features. `name = nothing` probes across all
+names (the name-less kwargs form with resolution/interval/feature filters)."""
 function infrastore_has_time_series(
     ::Type{T},
     owner::TimeSeriesOwners,
-    name::AbstractString;
+    name::Union{Nothing, AbstractString};
     resolution::Union{Nothing, Dates.Period} = nothing,
     interval::Union{Nothing, Dates.Period} = nothing,
     features...,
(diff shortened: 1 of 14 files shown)


src/infrastore.jl
```diff
@@ -1707,11 +1760,12 @@ _decode_element(element, encoding::RowEncoding) =
...
 """Route `has_time_series(owner, T, name; ...)` to the InfraStore store. Honors partial
 (subset) feature / resolution queries: matches if any stored series of type `T`
-contains at least the requested features."""
+contains at least the requested features. `name = nothing` probes across all
+names (the name-less kwargs form with resolution/interval/feature filters)."""
 function infrastore_has_time_series(
     ::Type{T},
     owner::TimeSeriesOwners,
-    name::AbstractString;
+    name::Union{Nothing, AbstractString};
```
jd-lara, line 1768: Is a union with nothing here, seems odd to me

src/infrastore.jl
```diff
@@ -1797,7 +1863,14 @@ function infrastore_has_any(owner; time_series_type::Union{Nothing, Type} = noth
     owner_id, _, category = _infrastore_owner_args(owner)
     types =
         time_series_type === nothing ? () : _infrastore_query_types(time_series_type)
```
jd-lara, line 1865: I don't know why claude always prefers the === instead of the is nothing()

jd-lara (approved):
seems ok to me, I am noticing that we have a lot of anonymous functions used in the code. AFAIK julia doesn't do great with those and compilation maybe that's an optimization for later

src/infrastore.jl
daniel-thom: The alternative is that name is a kwarg similar to resolution and interval. Based on usage at the call site, I think this makes sense.

src/infrastore.jl
daniel-thom: For many years, julia nerds demaned === because it was more efficient. While that is no longer true, the LLMs were probably trained on it.

src/infrastore.jl
jd-lara: but why can't we have a method to check on name and one that doesn't

daniel-thom: https://github.com/Sienna-Platform/InfrastructureSystems.jl/issues/613

PR #612, opened 2026-08-21
Draw component and supplemental attribute ids from one stream

daniel-thom:
Stacked on #611 — review that first; the diff here is one commit.

## What

`SystemData` carried two id counters, `next_component_id` and `next_supplemental_attribute_id`, so a component and a supplemental attribute could share a numeric id. This collapses them into a single `next_id`: an id now names exactly one object of either kind within a system.

- `get_next_component_id!` / `get_next_supplemental_attribute_id!` → one `get_next_id!(data)`.
- The two `assign_id!` methods had identical bodies once the counter merged, so they become one method over `Union{InfrastructureSystemsComponent, SupplementalAttribute}`. Semantics are unchanged: an `UNASSIGNED_ID` object gets the next id; one that already carries an id (deserialization, `assign_new_ids`) keeps it and the counter advances past it.
- `assign_new_id_internal!` draws from the shared counter.
- Serialization writes and reads one `"next_id"` key; `compare_values` skips the one counter unless `compare_uuids`.
- The 7-arg internal `SystemData` constructor loses a positional argument.

## Breaking

Deserializing a system written before this change now throws a `DataFormatError` naming the old keys, rather than defaulting the counter to 1. Defaulting would silently hand out ids that collide with the restored components — the silent-failure pattern the project bans. **Serialized systems on this line need regenerating, including any PSCB serialized cache.**

Downstream packages construct `SystemData` through the keyword constructor, so they are unaffected unless something calls the positional 7-arg form or the removed `get_next_*_id!` accessors. Worth a stack smoke test before merge.

## Verification

- Project formatter (`scripts/formatter/formatter_code.jl`): clean.
- Full suite: **8943 passing, 3 broken, 0 failing**. All 3 broken are the pre-existing `@test_broken false` markers in `test/test_time_series.jl`, unrelated to ids.
- `test/test_system_data.jl`: "Test independent integer id streams" → "Test single integer id stream", asserting attributes continue at 3/4 after two components and the next component gets 5.
- `test/test_serialization.jl`: one counter round-trips; the newly added component's id collides with neither the restored components nor the restored attribute.

https://claude.ai/code/session_018EAXp9EmbP5tyuuhkmL5AP

### DIFF
--- a/src/component.jl
+++ b/src/component.jl
@@ -5,7 +5,7 @@ subsystem membership sets.
 """
 function assign_new_id_internal!(data, component::InfrastructureSystemsComponent)
     old_id = get_id(component)
-    new_id = get_next_component_id!(data)
+    new_id = get_next_id!(data)
     mgr = get_time_series_manager(component)
     if !isnothing(mgr)
         InfraStore.replace_owner!(
--- a/src/internal.jl
+++ b/src/internal.jl
@@ -35,7 +35,8 @@ Internal storage common to [`InfrastructureSystemsType`](@ref)s.
 
 Components and supplemental attributes are identified by an integer `id` assigned by the
 owning [`SystemData`](@ref) when they are attached (see [`get_id`](@ref)); it is
-[`UNASSIGNED_ID`](@ref) until then. Each instance also holds optional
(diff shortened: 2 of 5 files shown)


jd-lara (approved):

jd-lara:
seems straight forward

PR #618, opened 2026-08-22
Refactor time series features as explicit dictionaries

daniel-thom:
## Summary
- replace time-series feature keyword splats with an explicit `features` keyword so unrelated or misspelled keywords cannot bleed into feature identity
- use `features::Union{Nothing, Dict} = nothing` on InfrastructureSystems hot paths, materializing `Dict{String, Any}` only when a stored key requires it
- reject `interval` for `SingleTimeSeries` and `NonSequentialTimeSeries` across retrieval, catalog queries, hashes, and removal
- update the benchmark harness, documentation, and regression coverage for the new interface
- temporarily source the coordinated [InfraStore branch](https://github.com/NatLabRockies/infrastore/tree/refactor/optional-time-series-features), where Julia APIs use `features::Union{Nothing, AbstractDict} = nothing` and forward absent features as `C_NULL`.

## Performance
Benchmark timings remain effectively unchanged. Compared with the explicit-empty-dictionary implementation, representative featureless operations allocate less memory per call:

- static and forecast reads: 640 fewer bytes/op
- time-series adds: 160–176 fewer bytes/op
- bulk removal: 528 fewer bytes/op

## Testing
- InfrastructureSystems full test suite: 9,234 passed
- coordinated benchmark run: `julia --project=. benchmark/bench.jl`
- InfraStore Julia feature-related tests passed; the remainder of its local suite requires a freshly built FFI library because the installed 0.8 artifact predates the `is_empty` symbol on `main`

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -33,6 +33,7 @@ UUIDs = "cf7118a7-6976-5b1a-9a39-7adc72f591a4"
 YAML = "ddb6d928-2868-570f-bddf-ab3f9cf99eb6"
 
 [sources]
+InfraStore = {url = "https://github.com/NatLabRockies/infrastore.git", rev = "refactor/optional-time-series-features", subdir = "julia/InfraStore.jl"}
 # Git rev, not a sibling path: a relative path resolves only in a workspace checkout and
 # fails outright in CI, which clones this repo alone. Pinned to a branch rather than a
 # version because PowerOpenAPIModels is not registered.
--- a/Project.toml
+++ b/Project.toml
@@ -33,7 +33,6 @@ UUIDs = "cf7118a7-6976-5b1a-9a39-7adc72f591a4"
 YAML = "ddb6d928-2868-570f-bddf-ab3f9cf99eb6"
 
 [sources]
-InfraStore = {url = "https://github.com/NatLabRockies/infrastore.git", rev = "refactor/optional-time-series-features", subdir = "julia/InfraStore.jl"}
 # Git rev, not a sibling path: a relative path resolves only in a workspace checkout and
 # fails outright in CI, which clones this repo alone. Pinned to a branch rather than a
 # version because PowerOpenAPIModels is not registered.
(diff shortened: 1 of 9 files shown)


jd-lara:
How much do we expect this one to break downstream applications?

daniel-thom:
> How much do we expect this one to break downstream applications?

This is absolutely a breaking change. We discussed it ~2 weeks ago and agreed that the impact (minimal because we don't think people are using features) was acceptable. I would be ok with revisiting the choice.

jd-lara:
> > How much do we expect this one to break downstream applications?
> 
> This is absolutely a breaking change. We discussed it ~2 weeks ago and agreed that the impact (minimal because we don't think people are using features) was acceptable. I would be ok with revisiting the choice.

we can evaluate the impact but I am supportive generaly of being more opinionated

jd-lara (commented):
My main concern with this is making the time series properties, interval, resolution etc a feature that is more generic. 

Mixing the fundamentals needed for a successful model and the tags to indentify the time series is kinda awkard

daniel-thom:
> My main concern with this is making the time series properties, interval, resolution etc a feature that is more generic.
> 
> Mixing the fundamentals needed for a successful model and the tags to indentify the time series is kinda awkard

What is more intuitive for users?
```
add_time_series!(sys, gen, ts, scenario = "high")
get_time_series(SingleTimeSeries, gen, "max_active_power", scenario = "high")
```
```
add_time_series!(sys, gen, ts, features = Dict(scenario => "high"))
get_time_series(SingleTimeSeries, gen, "max_active_power", features = Dict(scenario => "high"))
```
I now prefer the second because there is no ambiguity about what is a valid kwarg for get_time_series. Consider this case with a misspelled kwarg:
```
has_time_series(SingleTimeSeries, gen, "max_active_power", inteval = Hour(1))
```
It will return false instead of throwing an exception.

Consider one additional point: we should use time series keys more:
```
key = add_time_series!(sys, gen, ts, features = Dict(scenario => "high")). This means you only deal with features at add time.
get_time_series(gen, key)
```

PR #620, opened 2026-08-24
Fold has_time_series type translation to compile time

### LINKED ISSUE #613: Refactor has_time_series
The has_time_series code path in the new infrastore branch has historical baggage and performance issues. Inspect the methods holistically and refactor.

luke-kiernan:
Closes #613.

Rebased onto `IS4`, including #623 — the features refactor lands cleanly under this and
actually simplifies it: with `features::Union{Nothing, Dict}` replacing the kwargs
splat, the fast-route dispatch below keys on a plain `::Nothing` instead of an empty
`NamedTuple`. The base was also retargeted from `feat/infrastore-integration`, which is
merged into IS4.

Three commits: the backend type-translation work, the consolidation the issue asks for, then
the review fix making the query type required.

### 1. Type translation folds to compile time

`_infrastore_query_types` answers a pure function of the query type, but the `@eval`
table only baked constants for ten hand-listed `UnionAll`s. Parameterized concretes
(`SingleTimeSeries{Float64, 1}`) and `Union`s escaped it and stayed on the ~2 µs runtime
match loop. Neither is enumerable up front, and `Union`s are not dispatchable, so adding
methods could not close the gap — hence `@generated`. Verified with `code_typed` that
every spelling folds to a literal with a concrete return type:

| query type | returns |
|---|---|
| `SingleTimeSeries` | `(InfraStore.SingleTimeSeries,)` |
| `Forecast` | `(Deterministic, Probabilistic, Scenarios)` |
| `TimeSeriesData` | `AllStoredTypes()` |
| `SingleTimeSeries{Float64,1}` | `(InfraStore.SingleTimeSeries,)` |
| `Union{Deterministic, Probabilistic}` | `(Deterministic, Probabilistic)` |

`_infrastore_type_matches` and `_unparameterized_type` move up beside it — a generated
function may only call methods that already exist when it is defined. Other callers
unaffected; please don't move them back.

While it moved, `_infrastore_type_matches` lost its `row_type <: [some type]` return
values (review feedback). Both remaining answers are constants, and the type parameters
decide the subtyping:

```julia
_infrastore_matches_bound(::Type{<:B}, ::Type{B}) where {B} = true
_infrastore_matches_bound(::Type, ::Type) = false

_infrastore_type_matches(::Type{R}, ::Type{Q}) where {R <: TimeSeriesData, Q} =
    _infrastore_matches_bound(R, _infrastore_query_bound(Q))
```

What makes that work is moving the two irregularities out of the comparison and into a
normalization of the *query type*, where each answer is also a constant:

```julia
_infrastore_query_bound(::Type{T}) where {T} = _unparameterized_type(T)
# DST is `Deterministic`'s sibling, not its subtype, so widening to their common parent
# is how "a `Deterministic` query also matches a DST" is stated in the type domain
_infrastore_query_bound(::Type{<:Deterministic}) = AbstractDeterministic
```

A DST query is not widened, so it still matches DST alone. Verified against a
table-driven oracle of the previous semantics: parity across 15 query spellings × 6 row
types, no method ambiguity (Aqua agrees), and every match still folds to a constant.

**One deliberate behavior change**, from @copilot's review: `_unparameterized_type` now
normalizes a `Union` member-wise. It previously returned one unchanged, so
`Union{SingleTimeSeries{Float64, 1}, Probabilistic}` matched *no* rows at all — a silent
wrong answer, and a contradiction of the rule that a parameterized concrete narrows like
its UnionAll. Narrow spelling, but it was wrong rather than merely unsupported. Now
covered by a test.

Also on the probe path:

Also on the probe path:

- Thread the caller's already-resolved manager into `infrastore_has_any` /
  `infrastore_has_time_series` instead of resolving it a second time.
- `_infrastore_owner_id_category` at both probe sites — the probes never send the
  `owner_type` string `_infrastore_owner_args` builds.

### 2. The `has_time_series` family consolidates

Six definitions answered one question in four spellings, three of them forwarding to a
fourth. Three remain, and the query type is a **required** positional — per review, two
positional arguments with defaults is a shape to avoid, since reaching the second means
passing the first:

```julia
has_time_series(owner, ::Type{T}, name = nothing; resolution, interval, features)
has_time_series(owner;                            resolution, interval, features)
has_time_series(owner, name::AbstractString;      resolution, interval, features)
```

The only default left is a trailing one. Every call shape is preserved —
`has_time_series(owner)` alone is 33 call sites, 16 of them downstream — but none now
depends on a defaulted type argument.

**Gone: the type-first `has_time_series(T, owner[, name])`.** An IS3 leftover with no
caller in the workspace outside one line of IS's own serialization test — 145 owner-first
call sites against 1 type-first, and zero in PSY / PNM / POM / IOM / PSCB. It was added
in #357 and never picked up callers in either line. IS4 is breaking, so it is deleted
rather than deprecated. (Note that InfraStore.jl's own `has_time_series` is type-first
and unaffected; it is a different function taking a `Store`.)

**Newly reachable: a filter without a name**, e.g. `has_time_series(c; resolution = Hour(1))`.
The store has always served it and `infrastore_has_time_series` has always accepted
`name = nothing`; no public method passed it. Making the name optional alongside the
other narrowings is what removes the special case.

**The two-route fork is kept**, now as a dispatch on the query shape rather than two
public spellings:

```julia
_has_time_series(mgr, owner, ::Type{T}, ::Nothing, ::Nothing, ::Nothing, ::Nothing) =
    infrastore_has_any(...)          # owner-scoped probe
_has_time_series(mgr, owner, ::Type{T}, name, resolution, interval, features) =
    infrastore_has_time_series(...)  # catalog filter
```

It looks collapsible; it isn't. `has_for_owner` is consistently ~40% faster (848/861/871 ns
vs 1208/1213/1247 ns over three runs), so collapsing would regress the hotter shape. Tests
now pin the routing, so picking the wrong method is a failure rather than a silent
regression. Return-type inference stays `Bool` on every shape.

### On infrastore#55

Read it before choosing the shape here. It adds a derived surrogate `association_id` (a
pure hash of the full identity) plus an id-keyed metadata lookup. It does **not** collapse
this surface: `has_time_series` is a *partial* query — subset features, optional
resolution/interval, abstract query types — while the id needs a complete identity and a
single concrete stored type. What it will offer is a fast path for the fully-specified
case, and that case now funnels through one internal method where it can slot in. I don't
think this PR should wait on it, but happy to be told otherwise.

### Testing

- Full suite: **9277 pass, 2 broken** (pre-existing `@test_broken`), **0 fail**. Formatter clean.
- Heads-up for reviewers: `test/`'s manifest may still be on InfraStore v0.8.0.
  `Pkg.update("InfraStore")` in the test env if the narrowing testset errors.
- Downstream smoke test **not run** — there is no workspace-root project on my machine and
  the local PowerSystems checkout is pinned to a pre-InfraStore IS branch. Checked by
  inspection instead: no psy6 package calls the removed spellings, and every downstream
  extension (PNM's two, IOM's mock) is `(owner, T, name)`, which is unchanged. That is
  reasoning rather than verification.

### DIFF
--- a/src/infrastore.jl
+++ b/src/infrastore.jl
@@ -1809,11 +1809,14 @@ _decode_element(element, ::ScalarEncoding) = element
 _decode_element(element, encoding::RowEncoding) =
     _decode_static_values(reshape(element, 1, :), encoding, 1)[1]
 
-"""Route `has_time_series(owner, T, name; ...)` to the InfraStore store. Honors partial
-(subset) feature / resolution queries: matches if any stored series of type `T`
-contains at least the requested features. `name = nothing` probes across all names.
-`mgr` is the caller's already-resolved manager: every public entry point null-checks
-one before reaching here, so re-resolving it would be a second lookup per probe."""
+"""Route a narrowed `has_time_series` query to the InfraStore store — the general
+catalog filter, serving every query the owner-scoped probe in `infrastore_has_any`
+cannot. Honors partial (subset) feature / resolution queries: matches if any stored
+series of type `T` carries at least the requested features. Each of `name`,
+`resolution`, `interval`, and `features` is optional; `name = nothing` probes across
+all names. `mgr` is the caller's already-resolved manager: the public entry point
+null-checks one before reaching here, so re-resolving it would be a second lookup per
+probe."""
 function infrastore_has_time_series(
     ::Type{T},
     mgr::TimeSeriesManager,
--- a/src/infrastore.jl
+++ b/src/infrastore.jl
@@ -1811,10 +1811,12 @@ _decode_element(element, encoding::RowEncoding) =
 
 """Route `has_time_series(owner, T, name; ...)` to the InfraStore store. Honors partial
 (subset) feature / resolution queries: matches if any stored series of type `T`
-contains at least the requested features. `name = nothing` probes across all
-names (the name-less kwargs form with resolution/interval/feature filters)."""
+contains at least the requested features. `name = nothing` probes across all names.
+`mgr` is the caller's already-resolved manager: every public entry point null-checks
+one before reaching here, so re-resolving it would be a second lookup per probe."""
 function infrastore_has_time_series(
     ::Type{T},
+    mgr::TimeSeriesManager,
     owner::TimeSeriesOwners,
     name::Union{Nothing, AbstractString};
     resolution::Union{Nothing, Dates.Period} = nothing,
@@ -1822,9 +1824,8 @@ function infrastore_has_time_series(
     features::Union{Nothing, Dict} = nothing,
 ) where {T <: TimeSeriesData}
     _check_interval_supported(T, interval)
-    mgr = get_time_series_manager(owner)
-    store = mgr.data_store
-    owner_id, _, category = _infrastore_owner_args(owner)
+    store = mgr.data_store::Store
+    owner_id, category = _infrastore_owner_id_category(owner)
     feats = _infrastore_features(features)
     # Pure existence probe — a covering-index `SELECT 1 ... LIMIT 1` in the store; nothing
     # is listed, hydrated, or marshaled, so this is safe in hot per-component loops. A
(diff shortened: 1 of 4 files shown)


src/infrastore.jl
```diff
@@ -1823,16 +1824,39 @@ struct NoStoredTypes end
...
+# Mirrors InfraStore's semantics: a `Deterministic` (or `AbstractDeterministic`) query
+# also matches a `DeterministicSingleTimeSeries` (which reads as a `Deterministic`),
+# while a `DeterministicSingleTimeSeries` query matches DST only. Defined here rather
+# than beside the other row helpers because `_infrastore_query_types` generates against
+# it: a generated function may only call methods that already exist when it is defined.
+_infrastore_type_matches(
+    row_type::Type, ::Type{<:DeterministicSingleTimeSeries},
+) = row_type <: DeterministicSingleTimeSeries
+
+_infrastore_type_matches(row_type::Type, ::Type{<:AbstractDeterministic}) =
```
luke-kiernan: Writing `row_type <: [other_type]` as a return value feels very non-Julian. Can this be written using multiple dispatch instead?

src/time_series_interface.jl
```diff
@@ -885,79 +885,88 @@ end
...
-    return infrastore_has_any(val; time_series_type = T)
-end
-
 function has_time_series(
-    val::TimeSeriesOwners,
-    ::Type{T},
-    name::AbstractString;
+    owner::TimeSeriesOwners,
+    ::Type{T} = TimeSeriesData,
+    name::Union{Nothing, AbstractString} = nothing;
```
daniel-thom, line 914: Why this change? I'm generally not a fan of two positional arguments with default values. If you want to specify a value for the second, you are forced to pass the first.

luke-kiernan: Fair. The no-type version isn't used anywhere so I just made that argument required

daniel-thom (approved):

PR #603, opened 2026-08-06
Split the units machinery out of ProductionVariableCostCurve; add LossCurve

luke-kiernan:
Supports PowerSystems.jl#1728, where loss curves on HVDC lines and converters are bare `ValueCurve`s carrying no units, leaving their axes ambiguous.

Carrying units isn't specific to costs, so that machinery is split out one level:

```
ValueCurveWithUnits{T, U}          # value_curve + units type parameter
├── ProductionVariableCostCurve    # adds vom_cost
│   ├── CostCurve
│   └── FuelCurve
└── LossCurve
```

The new supertype owns what both families share (units parameter, value-curve accessors, equality/hashing, serialization, display). `ProductionVariableCostCurve` keeps only what makes a curve a cost: `get_vom_cost` and the cost-specific field deserializers.

`LossCurve` is deliberately not under `ProductionVariableCostCurve`: `ThermalGenerationCost.variable`/`HydroGenerationCost.variable` are typed on it, and `get_vom_cost` is defined on it — a loss curve has no VOM cost.

No behavior or serialization change for `CostCurve`/`FuelCurve`; methods moved, none removed or altered.

Tests: 1462 pass. The `Test generated structs` failure is pre-existing on `IS4` (verified against a clean baseline, unrelated).

Pairs with a PowerSystems.jl PR against `psy6`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #621, opened 2026-08-24
Split the units machinery out of ProductionVariableCostCurve; add LossCurve

luke-kiernan:
Supports PowerSystems.jl#1728, where loss curves on HVDC lines and converters are bare `ValueCurve`s carrying no units, leaving their axes ambiguous.

Carrying units isn't specific to costs, so that machinery is split out one level:

```
ValueCurveWithUnits{T, U}          # value_curve + units type parameter
├── ProductionVariableCostCurve    # adds vom_cost
│   ├── CostCurve
│   └── FuelCurve
└── LossCurve
```

The new supertype owns what both families share (units parameter, value-curve accessors, equality/hashing, serialization, display). `ProductionVariableCostCurve` keeps only what makes a curve a cost: `get_vom_cost` and the cost-specific field deserializers.

`LossCurve` is deliberately not under `ProductionVariableCostCurve`: `ThermalGenerationCost.variable`/`HydroGenerationCost.variable` are typed on it, and `get_vom_cost` is defined on it — a loss curve has no VOM cost.

No behavior or serialization change for `CostCurve`/`FuelCurve`; methods moved, none removed or altered.

Replaces #603, which was based on a much older `IS4`. The Copilot `hash` feedback on that PR was addressed separately and has already merged into `IS4` (`hash_from_fields` now hashes the concrete type), so it is not repeated here.

Tests: full suite 9090 pass, 0 fail, 2 broken. Both broken are pre-existing `@test_broken false` in `test/test_time_series.jl`, untouched by this change. Formatter clean.

Pairs with a PowerSystems.jl PR against `psy6`.


luke-kiernan:
Quoting from PSY [#1728](https://github.com/Sienna-Platform/PowerSystems.jl/issues/1728#issuecomment-5144028756)

> I just realized that the y axis units on loss curves are ambiguous too! From disucssion with Jose, we'll assume going forward that **both axes have the same units**. So loss curves still need a type parameter, but the way it applies is different from cost curves (where only the x axis is power). I'll have to rethink the struct hierarchy then

In light of that I will close and re-open once I have a design that fits the above

PR #622, opened 2026-08-25
Jd/expose

jd-lara:


src/utils/test.jl
```diff
@@ -42,6 +42,39 @@ function SimpleTestComponent(; name, val, internal = InfrastructureSystemsIntern
     return SimpleTestComponent(name, val, internal)
 end
 
+"""
+A component that carries a [`TimeSeriesKey`](@ref) in a field, the way a
+time-series-backed cost curve does. It exists so tests can prove a key survives a whole
+system round trip: on the wire the field is nothing but the key's association id, which
+only the store that minted it can resolve back into a key.
+"""
+mutable struct TimeSeriesKeyTestComponent <: InfrastructureSystemsComponent
```
daniel-thom: We started defining new types in the test package. Refer to test/components.jl, for example. That prevents polluting the main package namespace.

src/infrastore.jl
```diff
@@ -479,6 +479,7 @@ function serialize_single!(
     units::Union{Nothing, AbstractString} = get_units(sts),
     quantity_kind::Union{Nothing, AbstractString} = get_quantity_kind(sts),
     unit_system::Union{Nothing, AbstractUnitSystem} = get_unit_system(sts),
+    association_id::Integer = 0,
```
daniel-thom: Consider leaving this as `kwargs...` and forwarding as passthrough so that we don't have to set a default.

PR #624, opened 2026-08-26
Support cost curve unit conversions and FuelCurve → CostCurve

### LINKED ISSUE #501: Support cost struct unit conversions
`CostCurve` and `FuelCurve` keep track of `power_units::UnitSystem`. Also, `FuelCurve` has a `fuel_cost`, which is sometimes a scalar. It would be useful to have built-in support for

1. Conversion of `CostCurve`s and `FuelCurve`s from one `UnitSystem` to another
2. Conversion from `FuelCurve` to `CostCurve` if the `fuel_cost` is scalar

as we are currently doing both of these things manually _ad hoc_.

For 1, an easy approach would involve a function that takes the `CostCurve`/`FuelCurve`, the desired units, and a component reference or some other way to look up the relevant conversion factors. That might be a `good first issue`. Depending on how https://github.com/NREL-Sienna/PowerSystems.jl/issues/1134 is resolved, a cleaner solution might be possible later.

For 2, the converter function (or overloaded constructor — `CostCurve(fuel_curve)`) is trivial, also a `good first issue`.

luke-kiernan:
Closes #501. The missing primitive was a horizontal scale — a power-unit change is exactly `f(x) -> f(rho * x)` — so `scale_x` is added at each layer (all four `FunctionData` shapes, then `ValueCurve`, where it is defined on the input-output function a curve represents rather than its raw `function_data`, so `IncrementalCurve`/`AverageRateCurve` correctly pick up a chain-rule factor on their stored y data), alongside `scale_y` as a named alias of `*`; on top of those, `convert_power_units(curve, to, system_base_power, device_base_power)` for `CostCurve`/`FuelCurve` reuses the existing `_cost_coeff_ratio` table that backs `convert_cost_coefficient` so no new conversion math is introduced, and `CostCurve(::FuelCurve)` multiplies the value curve through by a scalar `fuel_cost`, erroring on a time-series-backed fuel cost and on a nonzero `startup_fuel_offtake` that a `CostCurve` would otherwise silently drop. Note for reviewers: a stacked follow-up (loss curves, PowerSystems.jl#1728) changes `convert_power_units` to take the x-axis ratio directly instead of the two base powers, and removes `_cost_coeff_ratio` in favor of PowerSystems owning that arithmetic — so treat this signature as short-lived, and say the word if you would rather review the two together.

### DIFF
--- a/src/InfrastructureSystems.jl
+++ b/src/InfrastructureSystems.jl
@@ -175,7 +175,7 @@ using .RelativeUnits:
     unitful_variant
 # Names not exported from the submodule are pulled in explicitly so the
 # `IS._strip_units(...)` / `IS.convert_cost_coefficient(...)` call sites work.
-using .RelativeUnits: _strip_units, convert_cost_coefficient
+using .RelativeUnits: _strip_units, convert_cost_coefficient, _cost_coeff_ratio
 include("random_seed.jl")
 include("utils/timers.jl")
 include("utils/assert_op.jl")
--- a/src/function_data/function_data.jl
+++ b/src/function_data/function_data.jl
@@ -525,6 +525,66 @@ end
 "Left shift the `FunctionData` by a scalar: (f << c)(x) = (f >> -c)(x) = f(x + c)"
 Base.:<<(fd::FunctionData, c::Real) = fd >> -c
 
+# SCALE THE AXES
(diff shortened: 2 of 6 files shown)


luke-kiernan:
I changed my mind, easier to just do 1 PR for the 2 issues

jd-lara:
Why do we want this?

PR #625, opened 2026-08-26
Lk/loss curve units v2

luke-kiernan:
- Add a `CostCurve`-from-`FuelCurve` function.
- Add a `LossCurve{T, U}` struct: lone field is `value_curve::T`. `U` provides power units, for **both axes**.
- Implement horizontal scaling for `ValueCurve`
- Introduce an abstract supertype, `ValueCurveWithUnits`. Unit conversion is handled by defining `y_axis_power_dimension` via multiple-dispatch on subtypes: 1 for loss curves, 0 for cost curves. [x axis is power for both, so that one rescales the same.]
- `convert_power_units` accepts: curve, from units, to units, and ratio. We could pass separate device base and system base power instead of ratio, but then we'd need a dispatch table in IS that's basically the same as PSY's `_du_to_su_ratio`.


jd-lara (commented):
Do we need a change in the schemas for the new loss_curve?

luke-kiernan:
> Do we need a change in the schemas for the new loss_curve?

It's a new struct definition, so yes. [I'm not very familiar with the schema stuff but it certainly looks like we'd need to add stuff for it, for serialization and grid DB reasons.]

jd-lara:
> > Do we need a change in the schemas for the new loss_curve?
> 
> It's a new struct definition, so yes. [I'm not very familiar with the schema stuff but it certainly looks like we'd need to add stuff for it, for serialization and grid DB reasons.]

Then we need to add this to SiennaSchemas and the other packages. It will be a good chance to learn it.

jd-lara (approved):
I left a comment in https://github.com/Sienna-Platform/SiennaSchemas/pull/39 but I like the change we can merge when the Schemas are updated

PR #626, opened 2026-08-28
Build a time series key from the id the store minted

daniel-thom:
infrastore removed `next_association_id!`, and with it every way for a caller to
name a catalog id. Only the OpenAPI row import supplies one now; every add lets
the catalog assign and reports what it chose. The three staging sites reserved an
id before staging so the key they returned could carry it, and there is no longer
a mechanism behind that.

So the key is built after the write instead of before it. `_infrastore_stage_data!`
returns a `StagedKey` — every field the key needs bar the `association_id`, with
the concrete key type as a parameter so the built key's type is known from the
staged one alone — and `build_key` turns it into the real key once the bulk add
hands back the id the row was filed under. A key stays a value: immutable, never
patched after the fact.

That moves when a key exists, which is visible to callers. A direct
`add_time_series!(mgr, owner, ts)` writes on the spot and still returns the key.
The same call through an open transaction returns `nothing`, because the addition
is still buffered and has no id. A caller that wants the keys opens the block with
`collect_keys = true` and reads `added_keys(txn)` after the flush; the context
collects them as it flushes, auto-flushes included, and drops them on a rollback
along with the rows they name. A bulk ingest that never asks retains none, which
is why the collection is opt-in.

`_transaction_added_keys` exists for the same reason: the multi-component
`add_time_series!(data, components, ts)` cannot take its keys from the block's
return value, since they do not exist until `commit!` runs after the block.

`get_time_series_key` follows the store's renames. `get_time_series_metadata` is
`get_metadata_by_id`, which answers a miss with `nothing` rather than raising —
the store treats a stale reference as a question worth answering, and resolving
one is past that question, so the miss becomes an `ArgumentError` here. Catalog
rows spell the id `id`, not `association_id`.

## Testing

`julia --project=test test/runtests.jl`: 9332 passed, 0 failed, 0 errors, against
infrastore `dt/association-ids`.

https://claude.ai/code/session_01ABLNRwKQQhLuArGp6oJyE2

### DIFF
--- a/docs/src/dev_guide/time_series.md
+++ b/docs/src/dev_guide/time_series.md
@@ -117,6 +117,20 @@ inside `with_deserialization_store(store) do ... end`. `deserialize(::Type{Syste
 binds the store it just opened around its own work, but components are deserialized by the
 parent package, which must wrap that pass the same way.
 
+The store mints that id as it inserts the row, which is why a key exists only once the
+addition has been written. A direct `add_time_series!(sys, owner, ts)` writes on the spot and
+returns the key. An addition staged into a `time_series_transaction` is still buffered, so it
+has no id yet and the call returns `nothing`; open the block with `collect_keys = true` and
+read `added_keys(txn)` once it has flushed:
+
+```julia
+key = time_series_transaction(sys; collect_keys = true) do txn
+    add_time_series!(txn, component, ts)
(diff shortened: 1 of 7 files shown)


jd-lara (approved):
I like the approach with the StageKey, it is clean.

PR #627, opened 2026-08-29
Fix time series correctness gaps, and make the tests that were silently skipped run

daniel-thom:
Two commits.

## Fix time series correctness gaps found by review (`2fd0e77e8`)

A correctness review of the time series layer turned up a set of independent defects; each fix carries a regression test. Grouped by area:

- **Static series iteration and slicing** — N-D iteration walked column-major while `length` counted timesteps; slicing to an empty range built a series indexing into an empty timestamp vector; `when()` now returns a `NonSequentialTimeSeries` rather than claiming a regular grid it does not have.
- **Store round-trip** — scalar forecast windows were written as `f64` regardless of element type, so a `Deterministic{Int64}` came back widened. A `TimeSeriesKey` snapshots the series' shape; if the series was removed and re-added, window arithmetic ran against the stale shape. Both read paths now compare what came back against what the key promised.
- **Queries, owners, removal** — abstract query types resolve through the catalog; reads from a detached owner throw `ArgumentError` instead of `FieldError` on `nothing`; `get_time_series_hashes` no longer silently drops one owner category; a by-name `remove_time_series!` matching nothing errors, as the by-key form already did.
- **Other** — `copy_time_series!` keyed supplemental attributes with `get_name`; `fast_deepcopy_system(skip_time_series = true)` handed back a read-only store; `from_iso_8601` rounded off fractional seconds a binary float cannot hold; several `MethodError`/`FieldError` paths became actionable `ArgumentError`s.
- **Forecast window arithmetic** — There were two off-by-one errors in the repo from long ago. There is no impact because the two functions were unused. `index_to_initial_time` was off by one (index 1 returned the *second* window; the last was unreachable), `get_total_period` ran one interval past the end; the function is not called in any Sienna codebase. `get_window_common` reported the valid range instead of throwing on a non-boundary `start_time`. Sienna code always calls `get_window` with a `DateTime` instead of an integer, and so the correct logic in `get_window_common` always ran.
- `_check_forecast_data` measured `Probabilistic`/`Scenarios` windows with `length` rather than `size(x, 1)`

The full commit message enumerates every change.

## Open question
Claude thinks that the original handling of normalization factor for forecast data is incorrect. The original code scaled each window by that window's maximum. Claude changed the code to use the overall maximum. I left the change until we confirm. Claude also changed the code to not mutate the caller's dict, which is a good change.

## Make the parity check and the multi-interval tests run (`423afa577`)

Two testsets were passing without asserting anything.

**The `element_type` conformance vectors have never run — here or in CI.** `_element_type_vectors_path()` walked up from `Base.find_package("InfraStore")`, which only lands on `<repo>/conformance` in an infrastore *source* checkout. IS depends on the registered package, so the walk resolved to `~/.julia/packages/conformance/...` and both testsets took their `@test_skip` branch. `main-tests.yml` checks out nothing but IS, so the IS↔Rust element-type parity check has never actually executed.

The registered package does not ship the corpus, and fetching it at test time would put a moving file on the network path of every test run. So it is vendored at `test/conformance/`, pinned at infrastore `69dabce`, with its provenance and regeneration step in a README. A live copy still wins when one is present (`INFRASTORE_CONFORMANCE_DIR`, or a `dev`ed InfraStore.jl beside one), so an infrastore dev tests against the corpus they are editing; a new testset warns — rather than fails, since a live checkout may be mid-edit — when the vendored copy has drifted. A bad `INFRASTORE_CONFORMANCE_DIR` now errors instead of silently falling through to the walk.

**The multi-interval `@test_broken` scaffolding was stale.** It was written against a store uniqueness key that omitted `interval`; InfraStore 0.10 includes it, so the scaffolding only hid live assertions. One testset wrapped its *entire body* in `catch e; e isa ArgumentError || rethrow()`, which would have swallowed a genuine `ArgumentError` regression anywhere in ~80 lines of assertions.

## Verification

Full suite: **9,478 passing, 0 broken, 0 failing** — 26 conformance assertions now run where 0 ran before, and the 2 previously-broken multi-interval testsets pass outright. Formatter clean.

Note: the branch is 6 commits behind `IS4`.

https://claude.ai/code/session_016AHH5e4NyWqgHvXDfBkedy


src/forecasts.jl
```diff
@@ -138,10 +142,18 @@ function iterate_windows_common(forecast)
 end
 
 """
-Return the Dates.DateTime corresponding to an interval index.
+Return the Dates.DateTime corresponding to a 1-based interval index: `index = 1` maps to
+the forecast's initial timestamp and `index = get_count(forecast)` to the last window.
 """
 function index_to_initial_time(forecast::Forecast, index::Int)
-    return get_initial_timestamp(forecast) + get_interval(forecast) * index
+    count = get_count(forecast)
```
daniel-thom: This is a bad bug from Claude. My best guess is that I ran a multi-repo workflow from the rust repo and Claude used the wrong rules for Julia. Need to pay closer attention to this type of issue.

src/infrastore.jl
```diff
@@ -858,6 +909,21 @@ function _infrastore_read_single(
...
         resolution = resolution, features = feats, time_range = time_range)
     values = _decode_stored_values(sts.data, _element_encoding(sts.element_type))
+    # A key holds a snapshot of the series' shape. If the series was removed and re-added
+    # under the same name and features, the key still resolves but describes a different
+    # array — and the window arithmetic above was computed against the stale shape. Catch
+    # that here rather than returning silently wrong values.
+    if sts.initial_timestamp != start || size(values, 1) != n
+        throw(
+            ArgumentError(
+                "TimeSeriesKey is stale: the stored series for '$name' starts at " *
```
daniel-thom: The key may not be stale. The user may simply be passing an invalid start time. We should report the inconsistency only instead of assuming staleness.

src/time_series_normalization.jl
```diff
@@ -9,29 +9,75 @@ Types of normalization that can be applied to time series data.
...
+    isempty(data) && throw(ArgumentError("Forecast data cannot be empty"))
+    max_value = maximum(get_max_value(v) for v in values(data))
+    if iszero(max_value)
+        throw(
+            ArgumentError(
+                "normalization_factor = max with a max value of 0.0 is not supported",
+            ),
+        )
+    end
+    return _divide_windows(data, max_value)
```
daniel-thom, line 52: @jd-lara This is a behavior change I noted in the PR description. The original code used the max value of each window instead of the overall max. Which is correct? Note that using max per window causes a difference in behavior if someone uses normalization with DeterministicSingleTimeSeries, where one max overall would be used.

jd-lara: Should be the overall max always over the entire series

jd-lara (approved):

PR #632, opened 2026-09-03
Serialize a store as arrays plus a document, without the catalog

jd-lara:
A store is an HDF5 file plus a SQLite catalog, and `serialize` has always written both. For a writer whose own document already carries the catalog's association rows — PowerSystems ships a `system.json` beside a `time_series.h5` — the `.sqlite` is a third file nobody reads: every row of it is already in the document, down to each row's `association_id` and its `data_hash` pointer into the arrays.

This adds the three calls that make arrays-plus-document a complete artifact, and teaches the existing deserialize path to recognise one.

## What's new

`serialize_arrays(store, file_path)` writes the array half alone. Atomic, unlike `serialize`: one file, one rename.

`deserialize_arrays(file_path)` opens such an artifact and mints an empty catalog for the document's rows to be replayed into.

`import_time_series_association_rows!(store_or_data, json)` is that replay — the counterpart of `import_supplemental_attribute_association_rows!`, which was already here. Rows only: the document carries locators, never values, so every row must name an array the store already holds, and each row keeps the `association_id` the document recorded. An import that assigned fresh ids would leave every reference the document holds pointing at the wrong series.

## The change to an existing function

`open_deserialized_infrastore_store` now detects a source with no `.sqlite` beside it and mints a catalog rather than copying one that is not there. Two consequences worth reviewing:

Such a source is copied out and given a fresh catalog **`read_only` or not**. There is no catalog to open in place, and minting one into the caller's directory is exactly the mutation `read_only` exists to prevent — so honouring `read_only` here would mean writing into the artifact the flag is protecting. A catalogless source therefore always gets the working-copy treatment.

The `cp` of the array file moves above the branch, since both paths need it. The old code copied the `.h5` and then conditionally the `.sqlite`; it now copies the `.h5`, returns early when there is no catalog, and otherwise copies the `.sqlite` unconditionally — the `isfile` guard that used to sit on that second `cp` is now the branch itself.

## Dependency

**Blocked on NatLabRockies/infrastore#71**, which adds `persist_arrays!` and `open_store_without_catalog`. Those are unreleased, so `[compat] InfraStore = "0.11"` is deliberately left alone here — it needs bumping once that lands and ships, and per the line's policy no compat bump goes in before release.

## Verification

Exercised against a local InfraStore checkout with that branch:

- `serialize_arrays` writes exactly `["arrays.h5"]` — no `.sqlite`
- `open_deserialized_infrastore_store` on that source returns a store with a minted, empty catalog
- the same call with `read_only = true` succeeds rather than failing on the absent catalog
- `deserialize_arrays` opens it directly

`scripts/formatter` clean. The full IS suite was not run: its test environment pins InfraStore by git rev, so it resolves the released 0.11 rather than the branch these calls need, and the three new functions cannot exist there until #71 ships. End-to-end coverage is in the PowerSystems branch that consumes them, where `to_file`/`from_file` round-trips a System through a two-file bundle.


src/sienna_archive.jl
jd-lara, line 1: This machinery can be reused for the I/O in PSIP

jd-lara (commented):
One open question about the behavior of this PR is how much safety we want to include. Should we store the hash of the json files and check it isn't corrupted? Things like that are missing

test/test_time_series.jl
```diff
@@ -6739,12 +6739,13 @@ end
     @test IS.has_time_series(component2, IS.SingleTimeSeries, ts_name)
     @test IS.has_time_series(component2, IS.DeterministicSingleTimeSeries, ts_name)
 
-    # removing from one component is fine, as long as another still has the SingleTimeSeries
+    # the derived forecast must go before its backing SingleTimeSeries; once both are gone,
```
jd-lara, line 6742: @daniel-thom This a change introduced in Infrastore 0.12 that  now enforces the DST-before-SingleTimeSeries removal ordering at the store level for every removal path. Can you confirm this is a desired change?

jd-lara, line 6742: @daniel-thom This a change introduced in Infrastore 0.12 that  now enforces the DST-before-SingleTimeSeries removal ordering at the store level for every removal path. Can you confirm this is a desired change?

daniel-thom: Confirmed. The code in main should behave the same way.

PR #633, opened 2026-09-09
Archive as zip instead of .tar.gz

acostarelli:
https://github.com/Sienna-Platform/PowerSystems.jl/pull/1785#discussion_r3971134520


src/sienna_archive.jl
```diff
@@ -73,20 +78,24 @@ end
...
     end
     dir = mktempdir()
-    open(CodecZlib.GzipDecompressorStream, path, "r") do io
-        Tar.extract(io, dir)
+    open(path, "r") do io
+        archive = ZipArchives.ZipReader(Mmap.mmap(io))
+        for i in 1:ZipArchives.zip_nentries(archive)
+            name = ZipArchives.zip_name(archive, i)
+            ZipArchives.zip_openentry(archive, i) do member
+                open(joinpath(dir, name), "w") do io
```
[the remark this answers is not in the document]
acostarelli: Not sure this is a risk if we are only creating flat archives.

Project.toml
```diff
@@ -52,6 +51,7 @@ InteractiveUtils = "1"
 JSON = "^1.5"
 LinearAlgebra = "1"
 Logging = "1"
+Mmap = "1.11.0"
```
jd-lara: Can you explain where this gets used?

acostarelli: This was a Claude recommendation. It's used when extracting the archive line 95, to avoid loading the zip into memory which will be large if the HDFs are large. TBH I don't know how much it saves us since I think we have to read each member into memory sequentially on line 98. But IDK I didn't measure timing.

daniel-thom: It prevents loading the entire archive into memory, which is good. It's also in the stdlib, so no real dependency. However, it needs to be
```
Mmap = "1"
```

daniel-thom (approved):

PR #637, opened 2026-09-11
Fix #636: windows locks zip file

### LINKED ISSUE #636: Windows CI: mktempdir cleanup fails (ENOTEMPTY) after the Sienna zip archive tests
On Windows, both `mktempdir() do dir ... end` blocks in `test/test_sienna_archive.jl` (the archive round trip at line 11, and the "compresses everything but HDF5" testset) fail at temp-dir cleanup:

```
┌ Error: mktempdir cleanup
│   IOError: rm("C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\jl_wLIOX6"): directory not empty (ENOTEMPTY)
│   @ Base.Filesystem .\file.jl:947 (mktempdir)
│   @ test\test_sienna_archive.jl:11
```

The logged error trips the runner's `@test length(get_log_events(tracker, Logging.Error)) == 0` in `test/InfrastructureSystemsTests.jl:82`, so the Windows Test-CI job fails even though every testset passes. Linux and macOS are green.

`ENOTEMPTY` from `rm(...; recursive=true)` on Windows means a file under the temp dir was still open when the block exited: on Windows an open handle blocks deletion. The candidates are the zip reader or writer in `create_sienna_archive` / `extract_sienna_archive` (`ZipArchives`), or the file `extract_sienna_archive` writes into `dir`, if any handle is not closed before returning. A `close` (or `do`-block scoping) on the archive handle before the function returns should fix it; a `GC.gc()` in the test is not a fix.

First seen on the PR that introduced the zip archive, #633 (merged 2026-09-09), whose own Windows job failed the same way:
- https://github.com/Sienna-Platform/InfrastructureSystems.jl/actions/runs/34402293390/job/102636843278

Reproduced on `jd/openapi-1x-suite` (#635), which does not touch the archive code:
- https://github.com/Sienna-Platform/InfrastructureSystems.jl/actions/runs/34552194351/job/103117243845

acostarelli:
Waiting for CI before marking ready for merge

### DIFF
--- a/src/sienna_archive.jl
+++ b/src/sienna_archive.jl
@@ -91,8 +91,9 @@ function extract_sienna_archive(path::AbstractString)
         throw(DataFormatError("$path does not exist"))
     end
     dir = mktempdir()
-    open(path, "r") do io
-        archive = ZipArchives.ZipReader(Mmap.mmap(io))
+    bytes = open(Mmap.mmap, path)
+    try
+        archive = ZipArchives.ZipReader(bytes)
         for i in 1:ZipArchives.zip_nentries(archive)
             name = ZipArchives.zip_name(archive, i)
             ZipArchives.zip_openentry(archive, i) do member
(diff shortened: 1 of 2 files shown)


acostarelli:
Closing and reopening because CI isn't running?

PR #638, opened 2026-09-11
Fix windows locks zip file

### LINKED ISSUE #636: Windows CI: mktempdir cleanup fails (ENOTEMPTY) after the Sienna zip archive tests

acostarelli:
Issue #636


daniel-thom (commented):
The diagnosis is right — a live mapping is what keeps the file locked on Windows, and the mapping handle outlives the `open(...) do io` because `close(io)` only drops the file handle, not the section. But I don't think this fix works on the Julia versions this branch is about: **on Julia ≥ 1.11, `finalize(bytes)` is a silent no-op.**

Since the 1.11 `Memory` refactor, `Mmap.mmap` attaches the unmap finalizer to the array's *backing memory*, not to the array (`stdlib/v1.11+/Mmap/src/Mmap.jl:261`):

```julia
A = unsafe_wrap(Array, convert(Ptr{T}, ...), dims)
finalizer(A.ref.mem) do x          # <- not finalizer(A), as it was in 1.10
    ... munmap / UnmapViewOfFile + CloseHandle ...
end
```

`finalize(A)` runs the finalizers registered *on `A`*, and there are none. Verified on 1.13:

| call | then touch `A[1]` |
|---|---|
| `finalize(A)` (this PR) | reads fine, exits 0 → nothing was unmapped |
| `finalize(A.ref.mem)` | segfault → mapping actually released |

On 1.10 the call does work — that's the `finalizer(A)` era — which is the kind of split that passes locally and keeps failing on Windows CI.

The new test is nondeterministic rather than failing outright, too: `bytes` is unreachable once `extract_sienna_archive` returns, so an incidental GC before `rm(path)` will sometimes release the lock and make it green anyway.

### On the manual `finalize`

There's no way around it — `names(Mmap)` is `[:Mmap, :mmap]`; there's no do-block form, no `close`, no `munmap`. ZipArchives 2.x doesn't help either: v1's `ZipFileReader`/`zip_open_filereader` (which had real `close` semantics) is gone, and 2.6 only has the buffer-based `ZipReader(::AbstractVector{UInt8})`. So an explicit `finalize` is the only lever available. What would help is making it appear exactly once, behind a do-block, instead of inline at the call site:

```julia
@static if VERSION >= v"1.11"
    _unmap!(bytes::Array) = finalize(bytes.ref.mem)
else
    _unmap!(bytes::Array) = finalize(bytes)
end

"""
Memory-map `path` read-only, pass the bytes to `f`, and unmap before returning.

`Mmap` offers neither a do-block form nor a `close`: the unmap is a finalizer, and since
Julia 1.11 it lives on the array's backing `Memory`, so `finalize(bytes)` is a no-op.
Unmapping eagerly matters on Windows, where a live mapping keeps the file locked and
blocks deleting or overwriting the archive.

`bytes` must not escape `f` — it dangles once this returns.
"""
function _with_mmap(f::Function, path::AbstractString)
    bytes = Mmap.mmap(path)
    try
        return f(bytes)
    finally
        _unmap!(bytes)
    end
end
```

with the call site taking the shape the rest of the file already uses:

```julia
    dir = mktempdir()
    _with_mmap(path) do bytes
        archive = ZipArchives.ZipReader(bytes)
        for i in 1:ZipArchives.zip_nentries(archive)
            ...
        end
    end
    return dir
```

Two smaller notes:

- `open(Mmap.mmap, path)` is redundant — `Mmap.mmap(path)` has an `::AbstractString` method that does that `open(...) do` internally.
- The escape hazard is the part worth a comment: `ZipReader` holds a reference to the buffer, so any use of `archive` after the unmap is a use-after-free (segfault, not an exception). The scoping here happens to be correct; a helper with the warning attached is what keeps it correct.

Worth keeping the added `rm(path); @test !isfile(path)` either way — the enclosing `mktempdir() do` only `@warn`s when cleanup fails, so without the explicit `rm` there'd be nothing to fail on.

I prototyped the helper on 1.13, including the zero-byte-file case (`mmap` returns a plain array with no finalizer there, so `_unmap!` is harmlessly a no-op).

daniel-thom:
@acostarelli The comment above is from Claude and seems plausible to me. Push back if necessary.

Also, @jd-lara agreed to limit our Julia version to 1.11 going forward.

acostarelli:
@daniel-thom Yeah my Claude warned a similar thing, I'm waiting to be able to run the windows CI again to see what happens.

src/sienna_archive.jl
```diff
@@ -101,6 +102,9 @@ function extract_sienna_archive(path::AbstractString)
                 end
             end
         end
+    finally
+        # mmap locks files on Windows so tempdir won't cleanup
+        finalize(bytes.ref.mem)
```
daniel-thom: Can you add a comment acknowledging private access based on no public method for unmmap? This should be fixed in the future in the stdlib.

daniel-thom (approved):

PR #639, opened 2026-09-11
Route Sienna archive extraction to a caller-supplied directory

hannahchubin:
Addresses PSY issue 1792 https://github.com/Sienna-Platform/PowerSystems.jl/issues/1792

## Summary

- `extract_sienna_archive` now takes `directory` as a **required** keyword, replacing the internal `mktempdir()`. Callers own the extraction location and its lifetime — no more silent `/tmp` default that breaks on HPC nodes with small or RAM-backed `/tmp`.
- New `_check_archive_member_name` guard: rejects any non-flat member name (absolute paths, `..` traversal, or names with a path separator) before extraction. Archives are flat by construction, so this closes a path-traversal window if a `.sn` were built by a permissive zip tool.
- Includes an unrelated compat fix on this branch: `TimerOutputs = "1"` (introduced in `c0a5f9194` on 2026-08-22, "bump TO") has no resolvable version — TimerOutputs.jl's latest is 0.5.29 and there is no 1.x line. Restored to `^0.5` so the environment resolves. Please treat this as a bug fix or drop it in review if there's a coordinated bump in flight elsewhere.

Paired PR in PowerSystems.jl: `hc/fix_tmp_extraction_dir` → `jd/serialization_refactor` (updates `_from_file_sienna` to pass the new keyword).

## Test plan

- [x] `Test Sienna archive round trip` — updated to pass `directory =` explicitly, still passes.
- [x] `Test Sienna archive write guards` — still passes.
- [x] `Test Sienna archive rejects unsafe member names` — new: unit-tests the guard on 6 unsafe names and end-to-end verifies extraction rejects a crafted archive with a nested-path member.
- Ran locally with `include("test/InfrastructureSystemsTests.jl"); run_tests("Sienna archive")` — 23 passing, 0 failing.


src/sienna_archive.jl
```diff
@@ -83,24 +83,36 @@ end
 """
 $(TYPEDSIGNATURES)
 
-Extract the Sienna archive at `path` and return the directory holding its
-members, which persists until the Julia session ends.
+Extract the Sienna archive at `path` into `directory` and return it. The
+caller owns `directory` and its lifetime.
 """
-function extract_sienna_archive(path::AbstractString)
+function extract_sienna_archive(path::AbstractString; directory::AbstractString)
```
daniel-thom, line 89: Let's think through this.

1.  A required kwarg feels off in this case. If it's required, why not make it a positional argument?
2. The variable names `path` and `directory` are synonyms. It's better if the developer can tell what's required without reading the docstring. For example, `extract_sienna_archive(archive_file, destination_dir)`.
3. Can you elaborate on your choice of making the caller own the lifetime of the directory? Julia will automatically delete a directory created by `mktempdir` when the process exits. So, you could simplify the function by defaulting the destination directory to `nothing` and then use `mktempdir` in that case. In PowerSystems user workflows, they probably want the extracted archive to live for the duration of the process. But let me know if you're thinking differently.

jd-lara (approved):

src/sienna_archive.jl
jd-lara: I didn't see this comment but it does need to be addressed before merge @hannahchubin

PR #643, opened 2026-09-15
Replace the hand-rolled @scoped_enum with EnumX

### LINKED ISSUE #289: Consider using EnumX.jl
We could remove the `scoped_enum` macro from this package and use `EnumX.@enumx` instead. We would need to replace all usages across PowerSystems.jl and PowerSimulations.jl.

luke-kiernan:
Closes #289.

`@scoped_enum` reached its values through a `Base.getproperty(::Type{T}, ::Symbol)` overload, which supersedes `Base.getproperty(x::Type, f::Symbol)` and throws out every call site in the dependency tree that inferred `getproperty` on a non-concrete type.

Measured with SnoopCompile on Julia 1.12.6, `using InfrastructureSystems`, same environment before and after:

| | invalidated MethodInstances | worst tree |
|---|---|---|
| before | 20,470 | `getproperty(::Type{…}, ::Symbol)` — 29,851 + 3,218 children |
| after | **3,754** | no getproperty trees; worst is GeoInterface at 758 |

An isolated A/B confirms the attribution: the same enum defined without only that overload invalidates 0, as does `@enumx`. The macro's other Base overloads contribute nothing, since they dispatch solely on newly defined types.

## What changes

`@scoped_enum` is kept, now expanding to `EnumX.@enumx` pinned to `Int64`. **`Name` is a module and the enum type is `Name.T`**, which is what belongs in a type annotation. We could define `NameType` to alias `Name.T` in the macro definition, then use `NameType` downstream over the ugly-looking `Name.T`. But it's still a breaking change--no more using straight `Name` as the type.

EnumX provides less than the old macro, so three things are defined once against `EnumX.Enum` rather than per enum: construction from a name (what the JSON and OpenAPI paths deserialize through), `convert` from an integer, and `serialize`/`deserialize`.

The second commit restores a non-allocating `Base.string`. `Base.Enums`' default prints through an `IOBuffer` (128 bytes a call) where the old macro returned an interned literal, and `string` is on the serialization path. It has to be emitted per enum: a `@generated` method on `EnumX.Enum` runs in the world age where IS was compiled and
cannot see `instances` for an enum a downstream package defines later.

## What stays unchanged

Definition sites, every `Name.MEMBER` value use, and the serialized representation — `EnumX.Enum <: Base.Enum`, so `string(x)` still yields the bare name. No data migration. `instances`, equality, `isless`/sorting and dispatch are unchanged and at performance parity; member access went from 59ns/16 bytes to 1.1ns/0 bytes.

## Notes for review

- An enum value has no `.value` field any more — use `Integer(x)`.
- A missing member is an `UndefVarError`, not an `ErrorException`.
- `propertynames(Name)` now returns module bindings (adds `:T` and the module name, loses declaration order). Use `instances(Name.T)` to enumerate members.

Full suite passes: 9,953 tests.

## Downstream

PowerSystems is ready on `lk/enumx` there and pins this branch. PowerTableDataParser, PowerSystemCaseBuilder, PowerFlows, IOM and POM define or consume these enums and need the same `.T` treatment.


jd-lara (commented):
the .T syntax is because is a constructor?

luke-kiernan:
> the .T syntax is because is a constructor?

Good question. `T` is just the default--one can provide a different name. We could go with `Name.Name` (`ACBusTypes.ACBusTypes`) or `Name.[FixedString]` (perhaps `Kind`) instead. There's also the option for a non-prefixed alias, by having the macro generate `const ACBusTypesT = ACBusTypes.T`...though then we're cluttering the top-level scope.

Per AI:

> You cannot have the bare name `Fruit` be the type while `Fruit.APPLE` also resolves, unless `Fruit` is a type with a `getproperty` overload — which is exactly the thing we're trying to avoid. One name can't be both a module and a type. So once values live in a module namespace, the type needs a second name inside it.

jd-lara:
> > the .T syntax is because is a constructor?
> 
> Good question. `T` is just the default--one can provide a different name. We could go with `Name.Name` (`ACBusTypes.ACBusTypes`) or `Name.[FixedString]` (perhaps `Kind`) instead. There's also the option for a non-prefixed alias, by having the macro generate `const ACBusTypesT = ACBusTypes.T`...though then we're cluttering the top-level scope.
> 
> Per AI:
> 
> > You cannot have the bare name `Fruit` be the type while `Fruit.APPLE` also resolves, unless `Fruit` is a type with a `getproperty` overload — which is exactly the thing we're trying to avoid. One name can't be both a module and a type. So once values live in a module namespace, the type needs a second name inside it.

Yeah, lets choose something less cryptic than T.

src/utils/utils.jl
```diff
@@ -375,76 +385,75 @@ julia> @scoped_enum Fruit APPLE = 1 ORANGE = 2
...
+    )
+end
+
+"""
+Construct an enum value from its name, e.g. `Fruit.T(:APPLE)` or `Fruit.T("APPLE")`.
+
+`EnumX` provides only construction from the integer value; the JSON and OpenAPI paths
+deserialize through the name.
+"""
+function (::Type{T})(name::Union{Symbol, AbstractString}) where {T <: EnumX.Enum}
```
daniel-thom: Is this type piracy? Same for Base.convert below.

luke-kiernan: Good catch. I moved those two inside the macro so that we're defining them per `@scoped_enum` created--that way the functions only catch our types, not any generic `EnumX`.

Claude code mentions that this would've been caught by Aqua's piracy check, but unfortunately that's broken on Julia 1.12 at the moment.

luke-kiernan:
> Yeah, lets choose something less cryptic than T.

I switched it to `Value`. Not the most natural, but at least that doesn't collide other things: we use `Type` and `Kind` elsewhere.

